PR #352, opened 2018-03-16
UNI-39456 Motion T.* animated on timeline not exporting correctly

vkovec:
handle case where Motion T.* is animated

Haven't been able to figure out how to animate Motion T instead of position again, so wasn't able to test if the fix worked.


benoithudson:
Cleaning up our PRs -- can we update this one and see if it fixes anything today? Seems easy enough to throw in.

vkovec:
this will not completely fix the issue as we were accidentally exporting Motion T before, and since then have updated the exporter to be more explicit in what it exports. Made a new issue for this, will also make a new PR: https://imgspc.atlassian.net/browse/UT-3115

PR #444, opened 2018-08-03
Uni-55015 blendshape color/vertex unit-tests

alexismorin:
We can't do much with these until we have actual stuff to test but I guess that when we do we can hit the ground running.


benoithudson:
We added these tests or equivalent through other PRs.

PR #483, opened 2019-09-03
V2.0.3 preview.1

benoithudson:
UT-2751: Fix packaging for FBX exporter.


benoithudson:
> Just so I better understand: the .7z file is not in the repo but generated each time we package ?

Exactly. The cmake script produces the 7z file, so we can develop on the uncompressed files but ship a compressed one.

Originally it was because all files had to be imported as assets, which was a nuisance, so we zipped them up. Now there's the `~` convention so we wouldn't be importing them as assets, but I don't know what I'd break by changing it to not compress.

PR #490, opened 2019-12-18
Ut 2995 export anim curve tangents

vkovec:


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -1745,23 +1752,9 @@ internal static HashSet<float> GetKeyTimes(AnimationCurve[] animCurves)
         /// NOTE : This is a work in progress (WIP). We only export the key time and value on
         /// a Cubic curve using the default tangents.
         /// </summary>
-        internal void ExportAnimationKeys (AnimationCurve uniAnimCurve, FbxAnimCurve fbxAnimCurve, 
-            UnityToMayaConvertSceneHelper convertSceneHelper)
+        internal static void ExportAnimationKeys (AnimationCurve uniAnimCurve, FbxAnimCurve fbxAnimCurve, 
+            UnityToMayaConvertSceneHelper convertSceneHelper, string uniPropertyName)
```
BadSingleton: is it me or the uniPopertyName parameter is not being used anywhere?

vkovec: you're right I don't use it anymore, good catch

benoithudson:
Seems like the same set of failures as the other PR.

vkovec:
I think it's failing because I didn't update the com.autodesk.fbx dependency

PR #494, opened 2020-01-20
Ut 3147 add transfer anim to recorder

vkovec:


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -3508,6 +3482,28 @@ internal enum TransformExportType { Local, Global, Reset };
...
                     }
 
+                    foreach (var unityGo in revisedExportSet)
+                    {
+                        IExportData iData;
+                        if(!exportData.TryGetValue(unityGo, out iData))
+                        {
+                            continue;
+                        }
+                        var data = iData as AnimationOnlyExportData;
```
benoithudson: Either cast with `(AnimationOnlyExportData)iData` to throw right here, or test for null (if the 'as' doesn't work).

com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorder.cs
```diff
@@ -45,40 +45,25 @@ protected override void EndRecording(RecordingSession session)
 #endif
                 var root = ((AnimationInputSettings)aInput.settings).gameObject;
                 clip.name = "recorded_clip";
-                clip.legacy = true;
```
benoithudson: why don't we need this code anymore?

com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorderSettings.cs
```diff
@@ -22,6 +22,167 @@ public bool ExportGeometry
...
+                    return;
+                }
+                if (string.IsNullOrEmpty(m_animDestBindingId))
+                    m_animDestBindingId = GenerateBindingId();
+
+                SetBinding(m_animDestBindingId, value);
+            }
+        }
+
+        static Transform GetBinding(string id)
```
benoithudson: Please comment what this is for (it's for domain reload, right?)

com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorder.cs
vkovec: I needed it to be legacy before because the Animation component only accepts legacy animation clips. Don't need it now that I pass the clip directly to the exporter.

com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorderSettings.cs
benoithudson: You didn't say why this function exists -- why do we need binding IDs at all.

benoithudson: Oh wait you do a bit later, in GenerateBindingIDs. Maybe move that function to be the first of the bunch?

PR #496, opened 2020-01-22
Fix FBX shape naming for blendshapes

lazlo-bonin:
If the name of a FbxShape backing a blendshape channel is left as an empty string, multiple blend shapes on the same object fail to export properly. Only the last blendshape modifier is used, regardless of which is enabled. This fixes it.


benoithudson:
Thanks @lazlo-bonin ; @vkovec will review this.

Do you have a small sample file we can use in a regression test?

lazlo-bonin:
Unfortunately I don't have a small example file at hand.

I noticed the issue when trying to export character models with multiple blendshapes from the now unsupported/deprecated Morph Character System on the Unity Asset Store.

We tested the blendshapes functionality by importing in Maya from a Unity exported model with multiple blendhsapes. The blendshape "envelope" name was present, but all actual blendshape "channels" showed as "Shape 1", "Shape 2", etc. Modifying any slider for those blendshapes only applied the last n blendshape.

I haven't tested the blendshapes in a reimport back into Unity specifically, but the same issue could be present.

PR #497, opened 2020-01-29
Basic support for optional recorder.

benoithudson:
Without the recorder, I don't get an error. With it, I can record FBX animation.

Do we have any tests for the recorder beyond manual tests?


benoithudson:
Closing because this was supposed to go to the 3.x release branch.

PR #498, opened 2020-01-30
Basic support for optional recorder.

benoithudson:
Without the recorder, I don't get an error. With it, I can record FBX animation.

(This time, on the right branch)


com.unity.formats.fbx/Editor/Unity.Formats.Fbx.Editor.asmdef
```diff
@@ -17,6 +17,12 @@
     "precompiledReferences": [],
     "autoReferenced": true,
     "defineConstraints": [],
-    "versionDefines": [],
+    "versionDefines": [
+        {
+            "name": "com.unity.recorder",
```
vkovec: we don't need to specify a version of the recorder?

benoithudson:
Soft dependence isn't supported in 2018.3 and isn't backwards-compatible. So, no soft dependence.

PR #499, opened 2020-01-31
Update to 2019.3

benoithudson:
Also remove a job we weren't using, and make a virtual job from a job that was
doing nothing on its machine (fixing a warning, speeding up builds).

This is stuff I saw while working on the soft dependence.

### DIFF
--- a/.yamato/promotion.yml
+++ b/.yamato/promotion.yml
@@ -6,7 +6,7 @@
 # tests.
 #
 test_editors:
-  - version: 2019.1
+  - version: 2019.3
 test_platforms:
   - name: win
     type: Unity::VM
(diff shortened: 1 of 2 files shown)


benoithudson:
Where is the yamato?

benoithudson:
Test failures are due to deep-convert changes, we didn't get around to updating the tests fully.

.yamato/promotion.yml
```diff
@@ -38,26 +38,6 @@ promotion_test_{{ platform.name }}_{{ editor.version }}:
 {% endfor %}
 {% endfor %}
 
-promotion_test_trigger:
```
vkovec: is this the test that was run after running the unit tests on all platforms but before promoting? the promotion test on 2019.1 I think

benoithudson: The promotion test is that test.

The promotion test trigger is causing that test to run if you click on the promotion test trigger and queue it up. Which is pretty useless since I could have clicked on the promotion test and queued it up.

PR #505, opened 2020-03-31
UT-3313 don't reduce keyframes after recording and before exporting

vkovec:
- baking, reducing keyframes and baking again can create unnecessary errors/discrepancies in the exported curve.
- for the particular issue being fixed (UT-3313), the recorded curves originally had constant/linear tangents, which then became cubic after recording (as the recorder loses tangent information). This combined with keyframe reduction resulted in jitteryness where there should have been clean transitions between blocking positions.


com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorder.cs
```diff
@@ -39,7 +39,9 @@ protected override void EndRecording(RecordingSession session)
                 var clipName = absolutePath.Replace(FileNameGenerator.SanitizePath(Application.dataPath), "Assets");
                 
 #if UNITY_2018_3_OR_NEWER
-                aInput.GameObjectRecorder.SaveToClip(clip, settings.FrameRate);
+                var options = new Animations.CurveFilterOptions();
+                options.keyframeReduction = false;
+                aInput.GameObjectRecorder.SaveToClip(clip, settings.FrameRate, options);
```
benoithudson: Given 2018.3 is the oldest we support we don't need the #if. Unless that #if is supposed to read 2019.3

PR #509, opened 2020-04-08
UT-3373 add separate list of editors to test when publishing vs on a PR

vkovec:


BadSingleton:
The test trigger will start tests on win/mac 2018.4 and 2019.3

The publish task has a dependency on win/mac 2018.4, 2018.4, 2019.1, 2019.2, 2019.3 and 2020.1 test tasks

The test trigger will run on any PRs while the publish will run only on branches matching the regex.

The way it's done right now is that 2018.4 and 2019.3 will run tests twice. I'd rephrase the dependencies to remove redundant versions from publish_editors and add a dependency on test trigger to the publish task

PR #511, opened 2020-04-20
Ut 3356 add create export set option

vkovec:
image

### DIFF
--- a/Integrations/Autodesk/maya/scripts/unityCommands.mel
+++ b/Integrations/Autodesk/maya/scripts/unityCommands.mel
@@ -783,6 +783,13 @@ proc string createTextFieldWithLabel(string $label, string $parent)
 }
 
 proc createExportSetDialog(int $exportType){
+    $origSelection = `ls -sl`;
+    if(size($origSelection) <= 0){
+        // nothing selected
+        print ("Nothing selected");
+        return;
+    }
+
     $exportAnim = false;
     $exportAnimOnly = false;
     switch($exportType){
@@ -803,9 +810,27 @@ proc createExportSetDialog(int $exportType){
     string $container = `formLayout -numberOfDivisions 100`;
     
     string $mainOptions = `columnLayout -adjustableColumn true -p $container`;
+
+    // go through selection to find common namespace to set as default
+    string $commonNamespace = ":";
+    int $maxNamespaceCount = 0;
+    for($i = 0; $i < size($origSelection); $i++){
+        string $currNamespace = getObjectNamespace($origSelection[$i]);
+        while ($currNamespace != ":"){
+            // get contents of namespace
+            string $namespaceContents[] = `namespaceInfo -ls $currNamespace`;
+            string $intersection[] = getIntersection($origSelection, $namespaceContents);
+            if(size($intersection) > $maxNamespaceCount){
+                $commonNamespace = $currNamespace;
+                $maxNamespaceCount = size($intersection);
+            }
+            $currNamespace = `namespaceInfo -p $currNamespace`;
+        }
+    }
     
     // get namespace
     string $namespaceField = createTextFieldWithLabel("Unity FBX namespace", $mainOptions);
+    textField -e -text $commonNamespace $namespaceField;
     
     string $modelFilePath = " 0 0";
     if(!$exportAnimOnly){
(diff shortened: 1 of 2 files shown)


Integrations/Autodesk/maya/scripts/unityCommands.mel
```diff
@@ -68,7 +68,7 @@ proc storeAttribute(string $node, string $attr, string $attrValue){
 
 
 proc int setExists(string $setName){
-    return stringArrayContains($setName, `listSets -allSets`);
+    return stringArrayContains($setName, `ls -sets`);
```
vkovec, line 71: listSets -allSets was not giving the full name of the set (including namespaces)

Integrations/Autodesk/maya/scripts/unityCommands.mel
```diff
@@ -633,13 +668,277 @@ proc setupNewExportSet(string $exportPath, int $exportAnimOnly, string $selected
...
+        -columnWidth2 $labelSize $textFieldSize
+        -columnAlign2 "right" "left"
+        -p $parent;
+    
+    text -label $label;
+    string $textField = `textField -width $textFieldSize`;
+    return $textField;
+}
+
+proc string getCommonNamespace(string $origSelection[]){
```
BadSingleton, line 793: So if I get this right, there's a bunch of objects and we want to get the "root-est" namespace?

Integrations/Autodesk/maya/scripts/unityCommands.mel
BadSingleton: Classic MEL

Integrations/Autodesk/maya/scripts/unityCommands.mel
```diff
@@ -633,13 +668,277 @@ proc setupNewExportSet(string $exportPath, int $exportAnimOnly, string $selected
     
     // add to the targetNamespace
     // iterate over all selected objects and rename
-    for($object in $selectedObjects){
-        rename $object ($targetNamespace + ":" + $object);
+    if ($namespace != ":"){
```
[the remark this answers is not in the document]
vkovec: the primary reason for adding items to a namespace is to make sure nothing is clobbered on import (if there are other objects with the same name in the scene). For export it's to help the user organize the scene and also to allow them to specify the namespace they would like stripped on export.

Integrations/Autodesk/maya/scripts/unityCommands.mel
vkovec: yes, if you select a bunch of objects and most of them are in the same namespace, then take that one. So if you have:

test1:test2:sphere
test1:test3:cube
test1:test2:cylinder

If should find `test1` because all objects have that namespace in common. But if you have:

test1:test2:sphere
test1:test2:cube
test1:test2:cylinder

Then it will take `test1:test2`.

I'll add this as a comment to the function.

Integrations/Autodesk/maya/scripts/unityCommands.mel
```diff
@@ -633,13 +668,277 @@ proc setupNewExportSet(string $exportPath, int $exportAnimOnly, string $selected
...
+    // get namespace
+    string $namespaceField = createTextFieldWithLabel("Namespace", $mainOptions, $labelSize, $textFieldSize);
+    textField -e -text $commonNamespace $namespaceField;
+    
+    string $modelFilePath = " 0 0";
+    if(!$exportAnimOnly){
+        string $modelFilePathField = createFilePathField("Model File Path", $mainOptions, $labelSize, $textFieldSize);
+        string $modelFileNameField = createTextFieldWithLabel("Model File Name", $mainOptions, $labelSize, $textFieldSize);
+        textField -e -text ($modelFilename + ".fbx") $modelFileNameField;
+        $modelFilePath = " " + $modelFilePathField + " " + $modelFileNameField;
```
benoithudson, line 913: have you tested paths with spaces?

vkovec: I haven't good idea. It shouldn't be an issue here though because I am passing the UI's ID not the path itself.

vkovec: @benoithudson just tested and it works with spaces.

benoithudson: Oh, I see, you're passing the field itself, not the value of the field. Makes sense!

PR #513, opened 2020-04-21
Ut 3377 update namespace in referenced export set

vkovec:
When you import a Maya file as a reference, it adds the contents of the file into a namespace that you can edit after import.

If the referenced file contains any export sets, then the namespace on these sets will no longer be correct as everything will be added to a new parent namespace.

e.g.
Say I have test.mb with the following contents:
```
n1:cube
n1:cube2
cube_exportSet -> namespace: n1
```
Now if I import it as a reference into a new scene it will import as:
```
test:n1:cube
test:n1:cube2
test:cube_exportSet -> namespace: n1
```
The namespace on cube_exportSet should then update to be `test:n1`

### DIFF
--- a/Integrations/Autodesk/maya/scripts/unityCommands.mel
+++ b/Integrations/Autodesk/maya/scripts/unityCommands.mel
@@ -16,13 +16,7 @@ global int $UnityExportAnim = 0;
 global int $UnityExportModel = 1;
 global int $UnityExportModelAnim = 2;
 
-global proc unityRemoveNativeMenuOnLoad(){
-    $removeSendToUnityMenu = `optionVar -q "UnityFbxForMaya_removeSendToUnityMenu"`;
-    if($removeSendToUnityMenu && `menu -exists "sendToUnityMenu"`){
-        //Remove the GamePipeline 'SendToUnity' button
-        menu -e -visible false "sendToUnityMenu";
-    }
-}
+// ==================== General Helpers ==============================
 
 // Load a specified settings file
 proc int loadUnityFbxSettings(string $fileName, string $settingType){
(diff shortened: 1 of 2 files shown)


Integrations/Autodesk/maya/scripts/unityCommands.mel
```diff
@@ -1028,4 +1050,35 @@ global proc unityExportModelAnim(){
 global proc unityCreateExportSet(){
     global int $UnityExportModelAnim;
     createExportSetDialog($UnityExportModelAnim);
+}
+
+global proc unityUpdateReferencedSetNamespace(){
```
vkovec, line 1055: this is the only new thing I added in this file that isn't just rearranging functions

Integrations/Autodesk/maya/scripts/unityCommands.mel
```diff
@@ -1028,4 +1050,35 @@ global proc unityExportModelAnim(){
...
+            continue;
+        }
+        
+        // There are three possibilities:
+        // 1. The namespace on the set is still the correct one (setNamespace)
+        // 2. The reference was imported for the first time and added to a new namespace (referenceNamespace + setNamespace)
+        // 3. The reference namespace was changed and the reference reimported (referenceNamespace + parsedNamespace)
+        // Simplest is to find the most common namespace in the set and update the attribute
+        string $unitySetContents[] = `listConnections $currSet`;
+        string $commonNamespace = getCommonNamespace($unitySetContents);
```
vkovec, line 1080: I'm not sure about this. Should I try to do something fancier to figure out whether the namespace should be 1,2, or 3?

The thing is I don't really have a good way to determine if this is the first time this set is being imported, or if the user changed the namespace and reloaded the file, or if the namespace on the set was determined by the user after import.

Also there is the possibility that not all of the set contents are in the same namespace.

vkovec:
Closing this for now as we have decided to replace the namespace field on the export set with a checkbox (to strip all namespaces), and therefore this scriptJob is no longer necessary

PR #514, opened 2020-04-22
UT-3382 map by ID instead of name for materials and meshes

vkovec:
- also force unique names for materials and textures

This fixes fogbug: https://fogbugz.unity3d.com/f/cases/1193945/

### DIFF
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -165,9 +165,9 @@ namespace UnityEditor.Formats.Fbx.Exporter
         Dictionary<FbxNode, Dictionary<FbxConstraint, System.Type>> MapConstrainedObjectToConstraints = new Dictionary<FbxNode, Dictionary<FbxConstraint, System.Type>>();
 
         /// <summary>
-        /// Map Unity material name to FBX material object
+        /// Map Unity material ID to FBX material object
         /// </summary>
-        Dictionary<string, FbxSurfaceMaterial> MaterialMap = new Dictionary<string, FbxSurfaceMaterial> ();
+        Dictionary<int, FbxSurfaceMaterial> MaterialMap = new Dictionary<int, FbxSurfaceMaterial> ();
 
         /// <summary>
         /// Map texture filename name to FBX texture object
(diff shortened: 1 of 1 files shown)


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -655,20 +668,23 @@ internal bool ExportMaterial (Material unityMaterial, FbxScene fbxScene, FbxNode
                 unityMaterial = DefaultMaterial;
             }
 
-            var unityName = unityMaterial.name;
-            if (MaterialMap.ContainsKey (unityName)) {
-                fbxNode.AddMaterial (MaterialMap [unityName]);
+            var unityID = unityMaterial.GetInstanceID();
+            if (MaterialMap.ContainsKey (unityID)) {
+                fbxNode.AddMaterial (MaterialMap [unityID]);
```
benoithudson, line 673: use TryGetValue so you only need to look it up once

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -655,20 +668,23 @@ internal bool ExportMaterial (Material unityMaterial, FbxScene fbxScene, FbxNode
...
 
+            fbxName = GetUniqueMaterialName(fbxName);
+
             if (Verbose) {
                 if (unityName != fbxName) {
-                    Debug.Log (string.Format ("exporting material {0} as {1}", unityName, fbxName));
+                    Debug.Log (string.Format ("exporting material {0} as {1}", unityID, fbxName));
                 } else {
-                    Debug.Log(string.Format("exporting material {0}", unityName));
+                    Debug.Log(string.Format("exporting material {0}", unityID));
```
benoithudson, line 687: I think the name is better here.

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -2515,18 +2536,58 @@ private void SetDefaultCamera (FbxScene fbxScene)
...
+        /// <param name="nameToIndexMap">The dictionary to use to map name to # of occurences</param>
+        private string GetUniqueName(string name, ref Dictionary<string, int> nameToIndexMap)
         {
             var uniqueName = name;
-            if (NameToIndexMap.ContainsKey (name)) {
-                uniqueName = string.Format (UniqueNameFormat, name, NameToIndexMap [name]);
-                NameToIndexMap [name]++;
-            } else {
-                NameToIndexMap [name] = 1;
+            if (nameToIndexMap.ContainsKey(name))
```
benoithudson, line 2543: Use TryGetValue here too, to reduce the number of lookups sometimes:
```
int index;
if (!nameToIndexMap.TryGetValue(name, ref index)
{
  index = 0;
}
else
{
  uniqueName = ...
}
nameToIndexMap[name] = index + 1;
return uniqueName;
```

benoithudson: and it's really `nameToCountMap` or `nameCounts`

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -2515,18 +2536,58 @@ private void SetDefaultCamera (FbxScene fbxScene)
         /// </summary>
         /// <returns>Unique name</returns>
         /// <param name="name">Name</param>
-        private string GetUniqueName(string name)
+        /// <param name="nameToIndexMap">The dictionary to use to map name to # of occurences</param>
+        private string GetUniqueName(string name, ref Dictionary<string, int> nameToIndexMap)
```
benoithudson, line 2540: doesn't need to be a ref I don't think; you're not assigning to nameToIndexMap

benoithudson: (if we were building lots and lots of tiny FbxExporters we might want to set the Dictionary to null in the constructor, and in GetUniqueName we'd allocate it if it was stil null. But we aren't)

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -1242,10 +1258,10 @@ private bool ExportInstance(GameObject unityGo, FbxNode fbxNode)
 
             FbxMesh fbxMesh = null;
 
-            if (!SharedMeshes.TryGetValue (unityPrefabParent.name, out fbxMesh))
+            if (!SharedMeshes.TryGetValue (unityPrefabParent.GetInstanceID(), out fbxMesh))
```
BadSingleton, line 1261: Is GetInstanceID consistent if a user closes and restarts Unity? If it is, then what happens if the Library folder is deleted. Does it still map?

vkovec: I might be misunderstanding but it only needs to be consistent for the duration of the export no? If the user closes and reopens Unity a new ID will be generated, however the map won't exist anymore.

BadSingleton: It was my misunderstanding, I thought we exported the Unity ID.

PR #515, opened 2020-04-23
UT-3389 export bones in original position instead of bindpose

vkovec:
When exporting bones of skinned meshes, their transform values are first set to the bindpose, so that the skin and bindpose are exported properly.

However, afterwards we need to set the bones back to their original transform values so that if the character is not in the bindpose on export, the same pose is maintained in the FBX.

This fixes fogbug: https://fogbugz.unity3d.com/f/cases/1185725/

### DIFF
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -957,11 +957,11 @@ namespace UnityEditor.Formats.Fbx.Exporter
                     fbxPreRotationQuaternion.Inverse();
 
                     // Multiply LclRotation by pre-rotation inverse to get the LclRotation without pre-rotation applied
-                    var fbxFinalQuat = fbxPreRotationQuaternion * EulerToQuaternion(new FbxVector4(fbxBone.LclRotation.Get()));
+                    var finalLclRotationQuat = fbxPreRotationQuaternion * EulerToQuaternion(new FbxVector4(fbxBone.LclRotation.Get()));
 
-                    var quatToEulerMatrix = new FbxAMatrix();
-                    quatToEulerMatrix.SetQ(fbxFinalQuat);
-                    fbxBone.LclRotation.Set(ToFbxDouble3(quatToEulerMatrix.GetR()));
+                    // Convert to Euler without axis conversion (Pre-rotation and LclRotation were already in Maya axis)
+                    // and update LclRotation
+                    fbxBone.LclRotation.Set(ToFbxDouble3(QuaternionToEuler(finalLclRotationQuat)));
                 }
             }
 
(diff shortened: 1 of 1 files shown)


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -938,6 +938,29 @@ SkinnedMeshRenderer unitySkin
...
+                foreach (var bone in bones)
+                {
+                    var fbxBone = MapUnityObjectToFbxNode[bone.gameObject];
+                    ExportTransform(bone, fbxBone, newCenter: Vector3.zero, TransformExportType.Local);
+
+                    // cancel out the pre-rotation from the exported rotation
+                    var fbxPreRotationEuler = fbxBone.GetPreRotation(FbxNode.EPivotSet.eSourcePivot);
+                    // Get the inverse of the prerotation
+                    var fbxPreRotationInverse = ModelExporter.EulerToQuaternion(fbxPreRotationEuler);
+                    fbxPreRotationInverse.Inverse();
```
BadSingleton, line 954: nit: That's a bit confusing. Maybe comment 1: "get the pre-rotation...", then comment 2: " ... then reverse it"
As it is now, I'm not sure if we get the inverse or invert the inverse

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -938,6 +938,31 @@ SkinnedMeshRenderer unitySkin
...
+                    var fbxPreRotationQuaternion = ModelExporter.EulerToQuaternion(fbxPreRotationEuler);
+                    // Inverse of the prerotation
+                    fbxPreRotationQuaternion.Inverse();
+
+                    // Multiply LclRotation by pre-rotation inverse to get the LclRotation without pre-rotation applied
+                    var fbxFinalQuat = fbxPreRotationQuaternion * EulerToQuaternion(new FbxVector4(fbxBone.LclRotation.Get()));
+
+                    var quatToEulerMatrix = new FbxAMatrix();
+                    quatToEulerMatrix.SetQ(fbxFinalQuat);
+                    fbxBone.LclRotation.Set(ToFbxDouble3(quatToEulerMatrix.GetR()));
```
benoithudson, line 964: does this take into account rotation order?

vkovec: By this point both the prerotation and local rotation used in the calculations have already been converted to Maya coords with XYZ rotation order. FBXAMatrix will return an euler in XYZ rotation order as well, so everything should already be in the right coordinates and rotation order.

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -938,6 +938,31 @@ SkinnedMeshRenderer unitySkin
...
+                {
+                    var fbxBone = MapUnityObjectToFbxNode[bone.gameObject];
+                    ExportTransform(bone, fbxBone, newCenter: Vector3.zero, TransformExportType.Local);
+
+                    // Cancel out the pre-rotation from the exported rotation
+
+                    // Get prerotation
+                    var fbxPreRotationEuler = fbxBone.GetPreRotation(FbxNode.EPivotSet.eSourcePivot);
+                    // Convert the prerotation to a Quaternion
+                    var fbxPreRotationQuaternion = ModelExporter.EulerToQuaternion(fbxPreRotationEuler);
```
benoithudson, line 955: looks like you don't need `ModelExporter.` here since you don't in the other calls

PR #517, opened 2020-04-24
Ut 3285 integration fails if settings file missing

vkovec:


Integrations/Autodesk/maya/scripts/unityCommands.mel
```diff
@@ -25,28 +68,66 @@ global proc unityRemoveNativeMenuOnLoad(){
...
+        // create file with default settings
+        $fileId = `fopen $fileName "w"`;
+        fprint $fileId $defaultSettings;
+        fclose $fileId;
+    }
+
+    // if the file still doesn't exist (failed to create)
+    // print an error message.
     if (`file -q -ex $fileName` == false){
         error ("Failed to find Unity Fbx "+$settingType+" Settings at: " + $fileName);
```
benoithudson, line 97: why not `eval($defaultSettings)` ?

maybe still leave a warning, but at least this way we'd have something working even in a broken installation

PR #522, opened 2020-05-13
UT-3390 fix failing unit tests on 2018.3

vkovec:
- pass optional import settings to ExportSkinnedMesh
- change import settings for both original and exported mesh to make sure they match

### DIFF
--- a/com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
+++ b/com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
@@ -541,7 +541,21 @@ namespace FbxExporter.UnitTests
             Assert.AreEqual (mesh.tangents, fbxMesh.tangents);
         }
 
-        private string ExportSkinnedMesh(string fileToExport, out SkinnedMeshRenderer originalSkinnedMesh, out SkinnedMeshRenderer exportedSkinnedMesh){
+        private delegate void SetImportSettings(ModelImporter importer);
+        private string ExportSkinnedMesh(
+            string fileToExport, 
+            out SkinnedMeshRenderer originalSkinnedMesh,
+            out SkinnedMeshRenderer exportedSkinnedMesh,
+            SetImportSettings setImportSettings = null)
+        {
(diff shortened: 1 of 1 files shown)


com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
```diff
@@ -852,7 +850,40 @@ public void TestBlendShapeExport(string fbxPath)
...
+                importer.importBlendShapeNormals = ModelImporterNormals.None;
+                importer.weldVertices = true;
+#else
+                importer.importBlendShapes = true;
+                importer.optimizeMesh = false;
+                importer.meshCompression = ModelImporterMeshCompression.Off;
+
+                // In 2018.3, the vertices still do not match unless no normals
+                // are imported.
+                importer.importNormals = ModelImporterNormals.None;
```
vkovec, line 876: in 2018.3 for blendshapes it is not sufficient to not import blendshape normals. There are discrepancies with the vertices if any normals are imported. (To be investigated with UT-3410 as well)

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
```diff
@@ -541,7 +541,21 @@ private void CompareMeshComponentAttributes(Mesh mesh, Mesh fbxMesh)
             Assert.AreEqual (mesh.tangents, fbxMesh.tangents);
         }
 
-        private string ExportSkinnedMesh(string fileToExport, out SkinnedMeshRenderer originalSkinnedMesh, out SkinnedMeshRenderer exportedSkinnedMesh){
+        private delegate void SetImportSettings(ModelImporter importer);
+        private string ExportSkinnedMesh(
+            string fileToExport, 
+            out SkinnedMeshRenderer originalSkinnedMesh,
```
BadSingleton, line 547: FYI: If 2018.3 has C#7+ support, you can return anonymous tuples: https://docs.microsoft.com/en-us/dotnet/csharp/tuples

PR #524, opened 2020-05-28
UT-3407 run promotion test on 2018.4 instead of 2019.3

vkovec:
- to not hit an issue with the timeline not being on the production server


vkovec:
this didn't actually fix the issue, and the test that fails will be removed/fixed

PR #530, opened 2020-07-16
UT-3484: When no objects are selected, exporting to Unity creates an empty FBX file

RajatGoswami:
Previously, when no objects are selected, exporting to Unity creates an empty FBX file:
image

In 3ds Max, export should only be possible if an object or export set is selected.

Selecting "Export to Unity" or "Export to Unity (Model Only)" with no selection now does nothing. This is to match the current behavior in Maya.

Updating ChangeLog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,5 @@
 # Changes in Fbx Exporter
 
-## [3.2.1-preview.1] - 2020-07-16
-### Changed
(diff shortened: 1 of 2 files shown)


RajatGoswami:
@vkovec It looks like the changelog can't be updated until the package version is also bumped. I've reverted my change to the changelog for now, we can consolidate all the changes in a future PR. Thanks for the review!

PR #531, opened 2020-07-17
UT-1945 Import Settings Reset Fix

nat-w:
When exporting an FBX and overwriting an existing one, the import settings and metadata can be preserved.
Added an option in the export settings to preserve import settings on an fbx that will be overwritten.


nat-w:
Screen Shot 2020-07-20 at 4 42 08 PM

New export menu.

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -3729,6 +3743,43 @@ private void ReplaceFile ()
...
+                var tempMetafilePath = Path.GetTempFileName();
+                
+                // get relative path for original fbx's metafile
+                var metafile = VersionControl.Provider.GetAssetByPath("Assets" + m_lastFilePath.Substring(Application.dataPath.Length)).metaPath;
+            
+                // save it to a temp file
+                try {
+                    File.Copy(metafile, tempMetafilePath, true);
+                } catch(IOException) {
+                    Debug.LogWarning (string.Format("Failed to copy file {0} to {1}", metafile, tempMetafilePath));
```
lassond, line 3759: Please add a little more infor, i.e. "The import settings for the FBX asset will be reset to the defaults"

com.unity.formats.fbx/Editor/ExportModelSettings.cs
```diff
@@ -95,6 +95,15 @@ public override void OnInspectorGUI ()
             exportSettings.SetExportUnredererd(EditorGUILayout.Toggle(exportSettings.ExportUnrendered));
             EditorGUI.EndDisabledGroup ();
             GUILayout.EndHorizontal ();
+            
+            GUILayout.BeginHorizontal();
+            EditorGUILayout.LabelField(new GUIContent("Preserve Import Settings for Existing FBX",
+                "If checked, the import settings from the overwritten FBX will be carried over to the new version."), GUILayout.Width(LabelWidth - FieldOffset));
+            // greyed out if animation only
+            EditorGUI.BeginDisabledGroup(exportSettings.ModelAnimIncludeOption == ExportSettings.Include.Anim);
```
lassond, line 103: Why is it disabled for the "animation only" case?

com.unity.formats.fbx/Editor/ExportModelSettings.cs
```diff
@@ -95,6 +95,15 @@ public override void OnInspectorGUI ()
             exportSettings.SetExportUnredererd(EditorGUILayout.Toggle(exportSettings.ExportUnrendered));
             EditorGUI.EndDisabledGroup ();
             GUILayout.EndHorizontal ();
+            
+            GUILayout.BeginHorizontal();
+            EditorGUILayout.LabelField(new GUIContent("Preserve Import Settings for Existing FBX",
```
lassond, line 100: This does not match the screenshot that's in the Conversation section of this PR (I only see "Preserve Import Settings"). Is it that the text is clipped because it is too long?

com.unity.formats.fbx/Editor/ExportModelSettings.cs
```diff
@@ -95,6 +95,15 @@ public override void OnInspectorGUI ()
...
+                "If checked, the import settings from the overwritten FBX will be carried over to the new version."), GUILayout.Width(LabelWidth - FieldOffset));
```
lassond, line 101: Is this the "hover over" text? As you can tell, I don't know much about the Unity GUI API :-)

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -3643,9 +3643,23 @@ internal enum TransformExportType { Local, Global, Reset };
                     Debug.LogWarning ("Export Cancelled");
                     return 0;
                 }
+
+                // make a temporary copy original metafile
```
lassond, line 3647: typo: "of the"

com.unity.formats.fbx/Editor/FbxExporter.cs
lassond: Here you should return ""

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -3729,6 +3743,43 @@ private void ReplaceFile ()
             }
         }
 
+        private string SaveMetafile()
+        {
+            // create temp file for original metafile
+            try {
```
lassond, line 3749: Why do you need this `try` statement?

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -3729,6 +3743,43 @@ private void ReplaceFile ()
...
+        {
+            // get relative path for new fbx's metafile
+            var fbxAsset = VersionControl.Provider.GetAssetByPath("Assets" + m_lastFilePath.Substring(Application.dataPath.Length));
+            var metafile = fbxAsset.metaPath;
+            
+            // replace metafile with original one in temp file
+            try {
+                File.Copy(metafilePath, metafile, true);
+            } catch(IOException) {
+                Debug.LogWarning (string.Format("Failed to copy file {0} to {1}", metafilePath, m_lastFilePath));
```
lassond, line 3779: Add the same "extra info" here (import settings will use default values)

com.unity.formats.fbx/Editor/ExportModelSettings.cs
nat-w: Yeah, the full message is clipped in the actual editor menu.

com.unity.formats.fbx/Editor/ExportModelSettings.cs
nat-w: Yes, it's a description that pops up when you hover over it.

TestProjects/FbxTests/ProjectSettings/ProjectVersion.txt
```diff
@@ -1,2 +1,2 @@
-m_EditorVersion: 2019.2.0a5
-m_EditorVersionWithRevision: 2019.2.0a5 (5ee2646333c0)
+m_EditorVersion: 2019.2.17f1
```
lassond, line 1: Looks like you are using an old version of Unity. Your should use the LTS version (Long Term Support), which is 2019.4 for Unity 2019.

Not worth changing for this PR but you should launch an install at some point

nat-w: Ah, I just realized I added some extraneous files. I'll remove them now.

com.unity.formats.fbx/Editor/ExportModelSettings.cs
```diff
@@ -95,6 +95,13 @@ public override void OnInspectorGUI ()
...
+            exportSettings.SetPreserveImportSettings(EditorGUILayout.Toggle(exportSettings.PreserveImportSettings));
+            EditorGUI.EndDisabledGroup();
```
lassond, line 103: What is this `EndDisabledGroup` matching against? 
I see you added a `BeginHorizontal` and `EndHorizontal` scope, but perhaps this one should be removed?

One nice way to organize these "scopes" is to add an indent to the code between them

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
```diff
@@ -1034,5 +1034,32 @@ public void TestBlendShapeExport(string fbxPath)
...
+            var cube = GameObject.CreatePrimitive(PrimitiveType.Cube);
+            var filename = GetRandomFbxFilePath();
+            ModelExporter.ExportObject(filename, cube);
+
+            // change an import setting
+            var importer = AssetImporter.GetAtPath(filename) as ModelImporter;
+            importer.importBlendShapes = false;
+            importer.SaveAndReimport();
+
+            // re-export with preserve import settings true and verify guid and settings are the same
```
lassond, line 1051: I don't see the guid comparison here (and below in the 2nd test)

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
```diff
@@ -1048,18 +1048,29 @@ public void TestPreserveImportSettings()
             importer.importBlendShapes = false;
             importer.SaveAndReimport();
 
-            // re-export with preserve import settings true and verify guid and settings are the same
+            // save original fbx's guid
+            var originalGuid = AssetDatabase.AssetPathToGUID(filename);
+
```
lassond, line 1053: This test will break if the default value for "importBlendShapes" changes in the future.

I think you should "remember" the original value like you do for the guid, then set the opposite value before exporting (!originalImportBlendShape)

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -3746,9 +3746,20 @@ private void ReplaceFile ()
         private string SaveMetafile()
         {
             var tempMetafilePath = Path.GetTempFileName();
-                
-            // check if path is relative or absolute
-            var fbxPath = AssetDatabase.LoadAssetAtPath(m_lastFilePath, typeof(Object)) != null ? m_lastFilePath : "Assets" + m_lastFilePath.Substring(Application.dataPath.Length);
+            
+            // Try as a relative path
```
lassond, line 3750: I think you got relative and absolute backwards (this is absolute, the next try is for a relative path)
Same goes for `ReplaceMetafile`

PR #534, opened 2020-07-29
UT-3125 Relative path of texture fix

nat-w:
Temp fbx file is now created in the directory where the exported fbx file will be created so that the relative path to textures is correct.

### DIFF
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -3448,7 +3448,8 @@ namespace UnityEditor.Formats.Fbx.Exporter
             {
                 // create a temp file in the same directory where the fbx will be exported
                 var exportDir = Path.GetDirectoryName(m_lastFilePath);
-                var tempFileName = Path.ChangeExtension(Path.GetRandomFileName(), ".fbx");
+                var lastFileName = Path.GetFileName(m_lastFilePath);
+                var tempFileName = Path.GetFileNameWithoutExtension(Path.GetRandomFileName()) + "_" + lastFileName;
                 m_tempFilePath = Path.Combine(new string[] { exportDir, tempFileName });
             }
             catch(IOException){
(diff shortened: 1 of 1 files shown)


nat-w:
Previously:
before

Now:
after

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -3438,17 +3438,23 @@ internal enum TransformExportType { Local, Global, Reset };
             Dictionary<GameObject, IExportData> exportData)
         {
             exportCancelled = false;
+            
+            m_lastFilePath = LastFilePath;
```
lassond, line 3442: Why did you move this statement?

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -3438,17 +3438,23 @@ internal enum TransformExportType { Local, Global, Reset };
...
             // This way we won't overwrite existing files.
-            try{
-                m_tempFilePath = Path.GetTempFileName();
+            try
+            {
+                // create a temp file in the same directory where the fbx will be exported
+                var exportDir = Path.GetDirectoryName(m_lastFilePath);
+                var tempFileName = Path.GetFileNameWithoutExtension(Path.GetRandomFileName());
+                // have to add .fbx to the end or else the temp file path won't match the file name
+                m_tempFilePath = exportDir + "/" + tempFileName + ".fbx";
```
lassond, line 3453: Here you should use `Path.Combine`. This will play better with other platforms

com.unity.formats.fbx/Editor/FbxExporter.cs
lassond: oh, I understand now, you use it below

com.unity.formats.fbx/Editor/FbxExporter.cs
lassond: I would also add the actual file name in m_tempFilePath. The metadata will at least give a hint about the right file name.

For example:
```
fileName = Path.GetFileName(m_lastFilePath);
m_tempFilePath = Path.Combine(exportDir, tempFileName + "_" + fileName);
```

PR #535, opened 2020-07-30
Add null checks when accessing bones.

nat-w:
Add null checks whenever accessing a bone, so that export doesn't fail with a broken rig.


nat-w:
A user from the forums had the issue with null bones, but I'm not sure what exactly caused it. I'm assuming either the bones are parented incorrectly or the rig is broken from import maybe.

PR #536, opened 2020-08-03
UT-3419 Duplicate meshes export as instances.

nat-w:
When exporting a hierarchy of game objects, any objects with duplicate meshes now export as instances.
Screen Shot 2020-08-03 at 1 57 12 PM

### DIFF
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -1313,7 +1313,7 @@ namespace UnityEditor.Formats.Fbx.Exporter
         }
 
         /// <summary>
-        /// if this game object is a model prefab then export with shared components
+        /// if this game object is a model prefab or the model has already been exported, then export with shared components
         /// </summary>
         [SecurityPermission(SecurityAction.LinkDemand)]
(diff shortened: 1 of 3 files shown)


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -1322,31 +1324,41 @@ private bool ExportInstance(GameObject unityGo, FbxScene fbxScene, FbxNode fbxNo
...
             }
-            
-            PrefabAssetType unityPrefabType = PrefabUtility.GetPrefabAssetType(unityGo);
-            // only export as an instance if the GameObject is part of a prefab instance
-            if (unityPrefabType == PrefabAssetType.MissingAsset ||
-                unityPrefabType == PrefabAssetType.NotAPrefab ||
-                PrefabUtility.GetPrefabInstanceStatus(unityGo) != PrefabInstanceStatus.Connected)
-            {
-                return false;
-            }
```
lassond, line 1333: Why don't you need these checks anymore?

nat-w: This check is checking if the object is a prefab, but now non-prefabs can also be exported as instances so it's not needed anymore. Later on, it'll return false if the object is neither a prefab nor a mesh that's been exported before.

PR #538, opened 2020-08-06
Fix label getting cut off.

nat-w:
Label Preserve Import Settings checkbox was getting cut off in a bad spot on Mac.
Screen Shot 2020-08-06 at 3 57 30 PM


nat-w:
The extra words in the label get cut off in Windows, but in Mac it looks a bit strange and resizing doesn't show the rest of the text.

PR #543, opened 2020-09-24
Ut 3732 fix errors when modifying fbx export settings preset

vkovec:
Errors from fogbug caused by presets creating new instances of Fbx Export Settings even though it is a singleton.
- Fix is to make it no longer a singleton but still keep an instance to the main settings (shown in the project settings)

This also fixes fogbug 1272666, compatible naming checkbox not aligned properly in UI.

Another issue is that the FBX export settings preset stores the values of the export/convert windows and overwrites them when the preset is applied without the users knowledge (as the values are not displayed in the preset).
Fixes related to this issue:
- display the export/convert window settings in the project settings
- when opening a new export/convert window, create a copy of the settings, defaulting to a default preset (if available), or the project settings.
- When modifying the settings in the export/convert window, the project settings are no longer updated, but the changes persist for the remainder of the Unity session.
- Disable buttons such as the "Install Integration" button when editing an FBX export setting preset
- rename "Show Convert UI" in project settings to "Display Options Window" and use it for both the Convert and Export windows instead of just the Convert window

TODO:
- If the export window settings diverge from the FBX export settings in the project settings, and the project settings are modified, then would this change overwrite the export window settings so they match the project settings again?
- When exporting without a window (Display Options Window disabled), and the file already exists what should happen (error, popup or increment file name)?
- Which settings should be used when exporting/converting without a window (the setting from the last opened window or the project settings)?

image


com.unity.formats.fbx/Editor/FbxExportSettings.cs
```diff
@@ -1462,7 +1467,7 @@ internal ScriptableSingleton()
         {
             if (s_Instance != null)
             {
-                Debug.LogError(typeof(T) + " already exists. Did you query the singleton in a constructor?");
+                // The user has most likely created a preset.
```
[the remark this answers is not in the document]
benoithudson: Feels like we should just have ExportSettings derive from ScriptableObject straight up, have the `instance` field as-is, but all the non-static functions should use `this`. Would that break everything?

com.unity.formats.fbx/Tests/FbxTests/FbxExportSettingsTest.cs
```diff
@@ -490,5 +491,21 @@ private void TestLocations(Dictionary<string,List<string>> data)
             }
         }
 
+        [Test]
+        public void TestFbxExportSettingsPreset()
+        {
+            var instance = ExportSettings.instance;
```
benoithudson, line 497: I think you need to test what happens if you make a second export setting as well.

benoithudson (changes requested):
This change is saying that ExportSettings is in fact not a singleton. If so then we must not be acting like it is.

com.unity.formats.fbx/Editor/FbxExportSettings.cs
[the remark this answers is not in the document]
vkovec: I went based off of this discussion on using presets with singletons on #devs-presets: https://unity.slack.com/archives/CQM4A947L/p1599134073006800

I can try modifying it.

com.unity.formats.fbx/Editor/FbxExportSettings.cs
```diff
@@ -182,8 +182,13 @@ static class Style
...
 
             if (GUI.changed) {
-                EditorUtility.SetDirty (exportSettings);
-                exportSettings.Save ();
+                // Only save the settings if we are in the Singleton instance.
+                // Otherwise, the user is editing a preset and we don't need to save.
+                if (this.targets.Length == 1 && this.target == ExportSettings.instance)
+                {
+                    EditorUtility.SetDirty(exportSettings);
+                    exportSettings.Save();
```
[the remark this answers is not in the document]
vkovec: I think it will save the ScriptableObject yes, but not the FbxExportSettings file on disk.

CHANGELOG.md
```diff
@@ -1,8 +1,16 @@
...
+### Changed
+- The Export and Convert to Prefab options can now be changed in the FBX Export Settings in Edit > Project Settings > Fbx Export.
+    - The Export and Convert to Prefab Options windows will use these settings by default.
+    - When modifying the settings in the Export/Convert to Prefab windows, the Project Settings are no longer updated, but the changes persist for the remainder of the Unity session.
+- Disable buttons such as the "Install Integration" button when editing an FBX export setting preset.
+- Rename "Show Convert UI" in project settings to "Display Options Window" and use it for both the Convert and Export windows instead of just the Convert to Prefab window.
+
 ### Fixed
 - Fix Export Model and Convert to Prefab Variant setting presets not serializing settings properly.
+- Fix NullReferenceException when creating a preset for the FBX export settings in Edit > Project Settings > Fbx Export.
```
benoithudson, line 13: - Fix error in an export when your project settings are not writeable (e.g. if you're using Perforce).

com.unity.formats.fbx/Documentation~/exporting.md
```diff
@@ -219,9 +219,14 @@ When exporting an FBX file, the following **Export Options** window opens, displ
...
 | __Export Unrendered__     | Check this option to export meshes that either don't have a renderer component, or that have a disabled renderer component. For example, a simplified mesh used as a Mesh collider. |
 |__Preserve Import Settings__ | Check this option to preserve all import settings applied to an existing fbx that will be overwritten in the export. If the GameObject is being exported as a new fbx, the import settings will not be carried over.|
+| __Don't ask me again__    | Check this option to use the same **Export Option** properties and hide this window when exporting to FBX in the future. You can reset this option by turning on the **Display Options Window** option under **Edit** > **Project Settings** > **Fbx Export** in Unity's top menu. |
```
benoithudson, line 222: You're usign double-stars here for the menus, but double-underscore later (in "exporting animation from the timeline")

com.unity.formats.fbx/Editor/ConvertToPrefabEditorWindow.cs
```diff
@@ -246,14 +247,121 @@ protected override bool Export()
...
+        protected override void StoreSettingsInSession()
+        {
+            base.StoreSettingsInSession();
+
+            // store Prefab Save Paths
+            if (m_prefabSavePaths == null)
+            {
+                return;
+            }
+            var joinedPaths = string.Join(",", m_prefabSavePaths);
```
benoithudson, line 273: comma is a valid character in a path; maybe choose semicolon (which is a common path separator) ?

benoithudson: Or, SetInt for the number and SetString for each index, creating a name for each one.
```
SettionState.SetInt(string.Format(SessionStoragePrefix, nameof(m_prefabSavePaths), n);
for(int i = 0; i < n; ++i) { SessionState.SetString($"....{i}", m_prefabSavePaths[i]); }
```

com.unity.formats.fbx/Editor/ExportModelEditorWindow.cs
```diff
@@ -101,6 +101,77 @@ protected float FbxExtLabelWidth
...
+        [SerializeField]
+        private List<string> m_fbxSavePaths;
+        internal List<string> FbxSavePaths
+        {
+            get
+            {
+                if (m_fbxSavePaths == null)
+                {
+                    // Try to restore from session, fall back to Fbx Export Settings
+                    var pathStr = SessionState.GetString(string.Format(SessionStoragePrefix, nameof(m_fbxSavePaths)), ExportSettings.instance.GetJoinedFbxSavePaths());
```
benoithudson, line 151: I think you've got this code twice, can you refactor it (and do the same fix of actually storing all the paths separately so you don't have trouble with funny-named paths)

com.unity.formats.fbx/Editor/ExportModelSettings.cs
```diff
@@ -1,25 +1,36 @@
 ﻿using UnityEditor;
 using UnityEngine;
+using System.Runtime.Serialization.Formatters.Binary;
+using System.IO;
```
benoithudson, line 4: Why did you need to add these (I'm probably just not seeing it)

com.unity.formats.fbx/Tests/FbxTests/FbxExportSettingsTest.cs
```diff
@@ -522,5 +538,204 @@ public void TestExportSettingsPresets()
...
+            instance.DisplayOptionsWindow = false;
+            Assert.That(instance.DisplayOptionsWindow, Is.False);
+
+            var preset = new Preset(ExportSettings.instance);
+            
+            instance.DisplayOptionsWindow = true;
+            Assert.That(instance.DisplayOptionsWindow, Is.True);
+
+            preset.ApplyTo(instance);
+            Assert.That(instance.DisplayOptionsWindow, Is.False);
```
benoithudson, line 556: also test that changing the preset doesn't change the instance

benoithudson (changes requested):
Looking good, just a few smaller comments principally about the storing a list of filenames.

com.unity.formats.fbx/Editor/ExportModelSettings.cs
vkovec: I was initially using a different method for copying the export settings, but forgot to remove these using statements.

CHANGELOG.md
vkovec: is there a fogbug or JIRA issue for this?

com.unity.formats.fbx/Editor/ExportModelEditorWindow.cs
```diff
@@ -101,6 +101,122 @@ protected float FbxExtLabelWidth
...
+            {
+                SessionState.SetString(string.Format(SessionStoragePrefix + "_{1}", varName, i), paths[i]);
+            }
+        }
+
+        protected void RestorePathsFromSession(string varName, List<string> defaultsPaths, out List<string> paths)
+        {
+            var n = SessionState.GetInt(string.Format(SessionStoragePrefix, varName), 0);
+            paths = new List<string>();
+            if (n > 0)
```
benoithudson, line 132: You don't need this if.

com.unity.formats.fbx/Editor/ExportModelEditorWindow.cs
```diff
@@ -101,6 +101,122 @@ protected float FbxExtLabelWidth
...
+                    if (!string.IsNullOrEmpty(path))
+                    {
+                        paths.Add(path);
+                    }
+                }
+            }
+
+            if (paths.Count <= 0)
+            {
+                paths = defaultsPaths;
```
benoithudson, line 146: This if can go to the top of the function, check if n == 0 then paths = defaultPaths, else paths = new List<string>

com.unity.formats.fbx/Editor/ExportModelSettings.cs
```diff
@@ -164,8 +188,25 @@ internal abstract class ExportOptionsSettingsSerializeBase : IExportOptions
...
+                return false;
+            }
+            return animatedSkinnedMesh == expOptions.animatedSkinnedMesh &&
+                mayaCompatibleNaming == expOptions.mayaCompatibleNaming &&
+                exportFormat == expOptions.exportFormat;
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
```
benoithudson, line 208: This is invalid: it can return different hash codes on two items that test equal.

Given you're basically got three bools, the hash code can be the bitmask of the three bools.

com.unity.formats.fbx/Editor/ExportModelSettings.cs
```diff
@@ -164,8 +188,25 @@ internal abstract class ExportOptionsSettingsSerializeBase : IExportOptions
...
+                return false;
+            }
+            return animatedSkinnedMesh == expOptions.animatedSkinnedMesh &&
+                mayaCompatibleNaming == expOptions.mayaCompatibleNaming &&
+                exportFormat == expOptions.exportFormat;
+        }
+
+        public override int GetHashCode()
+        {
+            return (animatedSkinnedMesh ? 1 : 0) | (mayaCompatibleNaming ? 1 : 0) << 1 | (int)exportFormat << 2;
```
benoithudson, line 208: can you add parens around the bitshifts? I want to make clear that you're shifting `(mayaCompatibleNaming ? 1 : 0)` and not `(animatedSkinnedMesh ? 1 : 0) | (mayaCompatibleNaming ? 1 : 0)`

com.unity.formats.fbx/Editor/ExportModelSettings.cs
```diff
@@ -190,7 +232,33 @@ internal class ExportModelSettingsSerialize : ExportOptionsSettingsSerializeBase
...
+        }
+
+        public override int GetHashCode()
+        {
+            var bitmask =  base.GetHashCode();
+            bitmask = (bitmask << 2) | (int)include;
+            bitmask = (bitmask << 2) | (int)lodLevel;
+            bitmask = (bitmask << 2) | (int)objectPosition;
+            bitmask = (bitmask << 1) | (exportUnrendered ? 1 : 0);
+            bitmask = (bitmask << 1) | (preserveImportSettings ? 1 : 0);
```
benoithudson, line 260: Probably better to use XOR than OR in case the enums get more values in them.

vkovec: not sure I understand. Wouldn't we still need to shift the bits more?

benoithudson: Ideally yes, since your data is so short. If you want to be sure to update the hash function then make a unit test that verifies your assumptions and has comments saying how and why.

But we don't need to depend on that; in general it's normal for hash codes to collide since you've only got 32 bits for the code, but potentially (though not in your case here) you've got way more data than you can fit in there. 

XOR mixes information about both operands, whereas with OR then if either operand is a 1, the other operand doesn't matter. So with XOR if you change any one field on its own, the hash value always changes, although if you change two at the same time then you can get a hash collision (eg. animatedSkinnedMesh off and include=5 versus animatedSkinnedMesh on and include=1).

Some bedtime reading about hash functions (slightly out of date now but still good basic info):
http://burtleburtle.net/bob/hash/index.html

benoithudson (approved):

PR #552, opened 2020-11-12
FBX-13 add doc sample unit test

vkovec:


com.unity.formats.fbx/Tests/FbxApiTests/SampleCodeTest.cs
```diff
@@ -0,0 +1,142 @@
...
+            if (Directory.Exists(m_tempFolder))
+            {
+                Directory.Delete(m_tempFolder, recursive: true);
+            }
+        }
+
+        // Export GameObjects sample function
+        public static void ExportGameObjects(Object[] objects)
+        {
+            string filePath = Path.Combine(Application.dataPath, "MyGame.fbx");
```
vkovec, line 38: It would be better to use a temp folder for this, but I copied this function exactly from the docs and thought it would be best to leave as is for most accurate testing.

benoithudson (approved):
Great!

Even greater would be if somehow we could embed the exact same source code file in but I don't know how we could do that nicely.

PR #554, opened 2020-12-01
Ut 3558 add build test

vkovec:


benoithudson (changes requested):
I could have sworn I reviewed this one already with the comments:
1. Make sure to have errors if a new editor platform gets added -- make those be `#elif` not `else`
2. use the same order of platforms for both constants.

benoithudson (approved):

benoithudson:
Can you send this to master? We should have no code changes on release/4.0 anymore.

PR #555, opened 2020-12-03
The documentation links changed and now we were pointing to the wrong…

benoithudson:
… page.

Fixes FBX-35.

### DIFF
--- a/com.unity.formats.fbx/Documentation~/index.md
+++ b/com.unity.formats.fbx/Documentation~/index.md
@@ -32,6 +32,6 @@ The Unity Integration for Autodesk® 3ds Max® feature supports the following ve
 
 ## Installing the FBX Exporter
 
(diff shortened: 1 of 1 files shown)


com.unity.formats.fbx/Documentation~/index.md
```diff
@@ -32,6 +32,6 @@ The Unity Integration for Autodesk® 3ds Max® feature supports the following ve
 
 ## Installing the FBX Exporter
 
-To install this package, follow the instructions in the [Package Manager documentation](https://docs.unity3d.com/Manual/upm-ui-install.html).
+To install this package, follow the instructions in the [Package Manager documentation](https://docs.unity3d.com/Manual/upm-ui-actions.html).
```
vkovec, line 35: the old links seems to point to installing from a registry. Isn't this the one we want as it is what most users will use? vs the more general actions menu.

benoithudson: Oh you're right, I misread that page as being all about scoped registries and other advanced stuff.

PR #558, opened 2020-12-17
FBX-45 add negative scale unit test

vkovec:
+ remove unused function in constraint test


com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
```diff
@@ -1115,5 +1115,64 @@ public void TestInstanceExport()
             GameObject fbxObj = AssetDatabase.LoadMainAssetAtPath(filename) as GameObject;
             Assert.AreEqual(fbxObj.transform.GetChild(0).GetComponent<MeshFilter>().sharedMesh.name, fbxObj.transform.GetChild(1).GetComponent<MeshFilter>().sharedMesh.name);
         }
+
+        private bool AreEqual(Vector3 a, Vector3 b, double epsilon = 0.0001)
```
benoithudson, line 1119: This is actually epsilon2 (for squared): epsilon^2 = 0.0001 means the vectors are equal to within a centimeter. Maybe bump this down to 1e-6 or even less?

benoithudson: Although I guess you aren't looking for rounding errors here but a very gross error, so it doesn't matter much.

benoithudson (commented):
Rename epsilon to epsilon2 and comment what it means in terms of meters. Other than that, this is good to merge.

PR #565, opened 2021-03-02
FBX-31 CameraVisitor.cs no longer using Reflection

RajatGoswami:
- `Handles.GetMainGameViewSize();` (public API) now used
- Modifying unit test to check for aspect ratio (set using the value returned by `GetSizeOfMainView()`)


com.unity.formats.fbx/Tests/FbxTests/FbxCameraTests.cs
```diff
@@ -102,6 +102,7 @@ public void PhysicalCameraTest()
 
             Assert.That( srcCam.fieldOfView, Is.EqualTo(origCam.fieldOfView).Within(EPSILON));
             Assert.That( srcCam.focalLength, Is.EqualTo(origCam.focalLength));
+            Assert.That( srcCam.aspect, Is.EqualTo(origCam.aspect));
```
RajatGoswami, line 105: PhysicalCameraTest attempts to export a physical camera. This calls into `ConfigurePhysicalCamera` in CameraVisitor.cs that executes `GetSizeOfMainView()` to calculate the fbxCamera's aspect ratio.

benoithudson (approved):

PR #574, opened 2021-03-15
FBX-30: FbxRecorderSettings uses reflection but doesn't need to

RajatGoswami:
* Adding public API call to ifdef
* Adding automated test file (TODO: currently only testing get/set binding functionality but eventually the entire Recorder workflow should be tested)
* Modifying the test assembly definition file to reference the Unity.Recorder assembly and set the `#if ENABLE_FBX_RECORDER` pre-processor definition

### DIFF
--- a/com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorderSettings.cs
+++ b/com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorderSettings.cs
@@ -91,12 +91,7 @@ namespace UnityEditor.Formats.Fbx.Exporter
         /// <returns>Transform associated with binding ID</returns>
         static Transform GetBinding(string id)
         {
-#if UNITY_2020_1_OR_NEWER
             return BindingManager.Get(id) as Transform;
-#else
-            var method = Type.GetType("UnityEditor.Recorder.BindingManager,Unity.Recorder.Editor").GetMethod("Get", System.Reflection.BindingFlags.Static | System.Reflection.BindingFlags.Public);
-            return method.Invoke(null, new object[] { id }) as Transform;
-#endif // UNITY_2020_1_OR_NEWER
         }
 
         /// <summary>
@@ -108,12 +103,7 @@ namespace UnityEditor.Formats.Fbx.Exporter
         /// <param name="obj">Unity Object</param>
         static void SetBinding(string id, UnityEngine.Object obj)
         {
-#if UNITY_2020_1_OR_NEWER
             BindingManager.Set(id, obj);
-#else
-            var method = Type.GetType("UnityEditor.Recorder.BindingManager,Unity.Recorder.Editor").GetMethod("Set", System.Reflection.BindingFlags.Static | System.Reflection.BindingFlags.Public);
-            method.Invoke(null, new object[] { id, obj });
-#endif // UNITY_2020_1_OR_NEWER
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


com.unity.formats.fbx/Tests/FbxTests/FbxRecorderTest.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    public class FbxRecorderTest : ExporterTestBase
+    {
+        public static string[] RecorderTestCases = new string[]
+        {
+            "Models/Player/Player.prefab"
+        };
+
+        [Test, TestCaseSource(typeof(FbxRecorderTest), "RecorderTestCases")]
+        public void TransferAnimationTest(string prefabPath)
```
vkovec, line 18: is this test specifically testing Get/SetBinding? please add a comment about it

com.unity.formats.fbx/Tests/FbxTests/FbxRecorderTest.cs
```diff
@@ -0,0 +1,42 @@
...
+            FbxRecorderSettings settings = ScriptableObject.CreateInstance(typeof(FbxRecorderSettings)) as FbxRecorderSettings;
+            settings.animationInputSettings = animationInputSettings;
+
+            settings.TransferAnimationSource = go.transform;
+
+            Transform skeletonNode = go.transform.GetChild(2);
+            settings.TransferAnimationDest = skeletonNode;
+
+            Assert.That(settings.TransferAnimationSource, Is.Not.Null);
+            Assert.That(settings.TransferAnimationDest, Is.Not.Null);
```
vkovec, line 38: if this test is more general, then maybe also add a check that Dest/Source have the right values (if possible).

com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorderSettings.cs
```diff
@@ -91,9 +91,12 @@ static string GenerateBindingId()
         /// <returns>Transform associated with binding ID</returns>
         static Transform GetBinding(string id)
         {
-            // return BindingManager.Get(m_BindingId) as GameObject;
+#if UNITY_2020_1_OR_NEWER
```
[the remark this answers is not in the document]
RajatGoswami: @benoithudson Do we know when BindingManager was made public so using reflection was no longer necessary? Is 2020.1 the appropriate year to check against in this case?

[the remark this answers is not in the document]
benoithudson: I don't, but as Viktoria mentions I think it would be the `com.unity.recorder` version not the Unity version. We can bump the required version (in `package.json`) for recorder to whatever shipped with Unity 2021.1 and it should have the API.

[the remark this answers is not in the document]
benoithudson: Check with the recorder team (Marie I guess?) what version would be most appropriate. Probably 2.5.something

com.unity.formats.fbx/Editor/Unity.Formats.Fbx.Editor.asmdef
```diff
@@ -20,7 +20,7 @@
     "versionDefines": [
         {
             "name": "com.unity.recorder",
-            "expression": "2.2.0-preview.4",
+            "expression": "2.5.5",
```
vkovec, line 23: please update the changelog and docs (if applicable) that we are updating the recorder requirement

RajatGoswami: Done!

benoithudson (approved):

vkovec (approved):

PR #575, opened 2021-04-20
FBX-32: Avoid the use of EditorClip

RajatGoswami:
The Timeline team does not have plans to make the EditorClip class public and unfortunately the TimelineClip class doesn't provide the same level of functionality.

The major difference is that TimelineClip does not inherit from `UnityEngine.Object` and so it not a "selectable" entity in the editor that can be inspected. This prevents us from using it to replace all instances of EditorClip in the exporter codebase.

The changes outlined in this PR are attempting to remove EditorClip where possible before we move to another system.


com.unity.formats.fbx/Editor/IExportData.cs
```diff
@@ -123,37 +123,25 @@ IExportOptions exportOptions
...
-
         /// <summary>
         /// Get the timeline clip from the given editor clip using reflection.
         /// </summary>
         /// <param name="editorClip"></param>
         /// <returns>the timeline clip or null if none</returns>
         private static TimelineClip GetTimelineClipFromEditorClip(object editorClip)
         {
-            object clip = GetPropertyReflection(editorClip, "clip");
+            object clip = editorClip.GetType().GetProperty("clip").GetValue(editorClip, null);
```
benoithudson, line 133: Do we have a unit test for this function? We should somehow make sure that an editorClip has a "clip" property.

com.unity.formats.fbx/Editor/IExportData.cs
```diff
@@ -123,37 +123,25 @@ IExportOptions exportOptions
...
-        /// <param name="editorClip"></param>
-        /// <returns>The GameObject bound to the editor clip or null if none.</returns>
-        private static GameObject GetGameObjectBoundToEditorClip(object editorClip)
+        /// <param name="timelineClip"></param>
+        /// <returns>The GameObject bound to the timeline clip or null if none.</returns>
+        private static GameObject GetGameObjectBoundToTimelineClip(TimelineClip timelineClip)
         {
-            var timelineClip = GetTimelineClipFromEditorClip(editorClip);
-            object parentTrack = timelineClip.parentTrack;
+            object parentTrack = timelineClip.GetParentTrack();
```
benoithudson, line 144:
```
	message: Library/PackageCache/com.unity.formats.fbx@e3e0e2667cb2-1618952766000/Editor/IExportData.cs(144,47): error CS1061: 'TimelineClip' does not contain a definition for 'GetParentTrack' and no accessible extension method 'GetParentTrack' accepting a first argument of type 'TimelineClip' could be found (are you missing a using directive or an assembly reference?)
```

benoithudson: Do we need to depend on a newer version of timeline?

benoithudson: Looks like we need to bump the dependency in package.json to 1.5.4 (which we can since we ditched support for Unity 2018).

vkovec: 1.5.4 seems to be broken in 2020.1

vkovec: 1.5.2 seems to be working though (from a quick test)

benoithudson: Ugh. Yeah, please check what the oldest timeline is that works on all versions.

benoithudson (approved):

PR #577, opened 2021-04-27
FBX-133 fix NullRefException when exporting timeline clip

vkovec:


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -3136,14 +3136,19 @@ int objectCount
...
         /// <param name="hierarchyToExportData">Map from GameObject hierarchy to animation export data.</param>
         [SecurityPermission(SecurityAction.LinkDemand)]
         internal int GetAnimOnlyHierarchyCount(Dictionary<GameObject, IExportData> hierarchyToExportData)
         {
             // including any parents of animated objects that are exported
             var completeExpSet = new HashSet<GameObject>();
-            foreach (var data in hierarchyToExportData.Values) {
+            foreach (var item in hierarchyToExportData) {
+                var data = item.Value;
+                if(data == null || data.Objects.Count <= 0)
```
benoithudson, line 3147: Is `data` a `UnityEngine.object`?

If so we should test it like a boolean rather than test for null, so we can also catch "missing" values.

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -3136,14 +3136,19 @@ int objectCount
...
+                {
+                    Debug.LogWarningFormat("No animation data to export for objects in hierarchy: {0}", item.Key);
```
benoithudson, line 3149: Do we need this warning?

It feels wrong on two levels:
- if I select a bunch of objects and some of them have anim but others don't, I shouldn't need to care about those that don't.
- the caller should be in charge of determining what to do about there's no animation data to export. We should just report `0` and be happy.

com.unity.formats.fbx/Editor/FbxExporter.cs
vkovec: data is an interface containing a HashSet of GameObjects:
```
 internal interface IExportData
 {
     HashSet<GameObject> Objects { get; }
 }
```

com.unity.formats.fbx/Editor/FbxExporter.cs
vkovec: yeah you're right maybe not necessary

PR #578, opened 2021-04-27
FBX-134 export skinned mesh if exporting blendshape animation

vkovec:
make sure to export the mesh and blendshapes in animation only export mode.

### DIFF
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -2877,8 +2877,7 @@ namespace UnityEditor.Formats.Fbx.Exporter
                         ExportLight (go, fbxScene, node);
                     } else if (compType == typeof(Camera)) {
                         ExportCamera (go, fbxScene, node);
-                    } else if (compType == typeof(SkinnedMeshRenderer))
-                    {
+                    } else if (compType == typeof(SkinnedMeshRenderer)) {
                         // export only what is necessary for exporting blendshape animation
                         var unitySkin = go.GetComponent<SkinnedMeshRenderer>();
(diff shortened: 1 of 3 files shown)


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -2877,6 +2877,12 @@ private FbxNode CreateFbxNode(GameObject unityGo, FbxScene fbxScene)
                         ExportLight (go, fbxScene, node);
                     } else if (compType == typeof(Camera)) {
                         ExportCamera (go, fbxScene, node);
+                    } else if (compType == typeof(SkinnedMeshRenderer))
```
benoithudson, line 2880: Either follow the convention in this function of
```
} else if (...) {
```
or follow the Unity convention of 
```
}
else if (...)
{
```
but not halfway!

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -2877,6 +2877,12 @@ private FbxNode CreateFbxNode(GameObject unityGo, FbxScene fbxScene)
...
+                    {
+                        // export only what is necessary for exporting blendshape animation
+                        var unitySkin = go.GetComponent<SkinnedMeshRenderer>();
+                        var meshInfo = new MeshInfo(unitySkin.sharedMesh, unitySkin.sharedMaterials);
```
benoithudson, line 2884: check for a missing `sharedMesh` here

benoithudson: unless ExportMesh already does

vkovec: there's an IsValid check in ExportMesh

PR #585, opened 2021-05-31
Add null material check

natalia-wh:
Fix null reference exception when materials are missing.
Default material is assigned if material is null.

Resulting fbx of gameobject will missing materials:

Screen Shot 2021-05-31 at 3 38 55 PM


vkovec (approved):
great!

PR #588, opened 2021-06-01
FBX-97 Fix error when exporting object with empty name

natalia-wh:
Fix ArrayOutOfBoundsException when trying to export a gameobject with no name.
Added to changelog and combined both unreleased sections into one.

### DIFF
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -4627,7 +4627,7 @@ namespace UnityEditor.Formats.Fbx.Exporter
         private static string ConvertToMayaCompatibleName(string name)
         {
             if (string.IsNullOrEmpty(name)) {
-                name = InvalidCharReplacement.ToString();
+                return InvalidCharReplacement.ToString();
             }
             string newName = RemoveDiacritics (name);
 
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -4628,7 +4628,7 @@ namespace UnityEditor.Formats.Fbx.Exporter
         {
             string newName = RemoveDiacritics (name);
 
-            if (char.IsDigit (newName [0])) {
+            if (newName.Length > 0 && char.IsDigit (newName [0])) {
                 newName = newName.Insert (0, InvalidCharReplacement.ToString());
             }
 
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -4626,9 +4626,12 @@ namespace UnityEditor.Formats.Fbx.Exporter
 
         private static string ConvertToMayaCompatibleName(string name)
         {
+            if (string.IsNullOrEmpty(name)) {
+                name = InvalidCharReplacement.ToString();
+            }
             string newName = RemoveDiacritics (name);
 
-            if (newName.Length > 0 && char.IsDigit (newName [0])) {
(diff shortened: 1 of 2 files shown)


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -4628,7 +4628,7 @@ private static string ConvertToMayaCompatibleName(string name)
         {
             string newName = RemoveDiacritics (name);
 
-            if (char.IsDigit (newName [0])) {
+            if (newName.Length > 0 && char.IsDigit (newName [0])) {
```
vkovec, line 4631: this could be string.IsNullOrEmpty() in case of null string (not sure if that's possible though)

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -4628,7 +4628,7 @@ private static string ConvertToMayaCompatibleName(string name)
         {
             string newName = RemoveDiacritics (name);
```
vkovec, line 4629: I think the empty check should be before this call to be safe

com.unity.formats.fbx/Editor/FbxExporter.cs
vkovec: also will this leave the name empty? wondering if we should treat empty as an invalid char and replace it with `InvalidCharReplacement` as well, what do you think?

natalia-wh: The bug was that if the name was empty, you would get the error while trying to rename the fbx file in the export window. It wouldn't let you rename the fbx file, and you would have to cancel the export. 
This leaves the name empty, but lets you rename the file in the export window. The exporter makes you choose a filename for export anyways, since I don't think a ".fbx" file would work anywhere.

com.unity.formats.fbx/Editor/FbxExporter.cs
natalia-wh: If the name is null or empty, should it exit or reassign something to the name? (Like `InvalidCharReplacement` or a default name maybe)

com.unity.formats.fbx/Editor/FbxExporter.cs
vkovec: I thought this would give the error even in a hierarchy where the root object exported has a name but the child does not. e.g.:

>GameObject
>> {empty name}

com.unity.formats.fbx/Editor/FbxExporter.cs
vkovec: If possible try to assign `InvalidCharReplacement`

com.unity.formats.fbx/Editor/FbxExporter.cs
natalia-wh: It doesn't give an error, but the model inside the fbx object has no name. It doesn't really seem to effect anything; I was able to import the fbx into blender fine (I don't have a copy of Maya).

empty

I could assign the replacement character as a default name for that case though.

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -4626,6 +4626,9 @@ private static string RemoveDiacritics(string text)
 
         private static string ConvertToMayaCompatibleName(string name)
         {
+            if (string.IsNullOrEmpty(name)) {
+                name = InvalidCharReplacement.ToString();
```
vkovec, line 4630: you can return the result here

vkovec (approved):
small comment otherwise looks good, will let you know what happens with the test

PR #591, opened 2021-06-02
FBX-143 fix zero scale causing issues in 3ds max

vkovec:


benoithudson (approved):
Ouch. But yeah, seems like a reasonable hack.

natalia-wh (approved):

PR #594, opened 2021-06-04
FBX-142 Fix reset settings

natalia-wh:
Add reset functionality for FBX export project settings menu.
Add `Reset()` to `FbxExportSettings`.
Fix `dccOptionNames` being reset to null instead of empty list.

### DIFF
--- a/com.unity.formats.fbx/Editor/FbxExportSettings.cs
+++ b/com.unity.formats.fbx/Editor/FbxExportSettings.cs
@@ -871,11 +871,11 @@ namespace UnityEditor.Formats.Fbx.Exporter {
             dccOptionPaths = new List<string>();
             dccOptionNames = new List<string>();
             BakeAnimationProperty = false;
-            ExportModelSettings = ScriptableObject.CreateInstance (typeof(ExportModelSettings)) as ExportModelSettings;
-            exportModelSettingsSerialize = ExportModelSettings.info;
+            exportModelSettingsSerialize = new ExportModelSettingsSerialize();
+            ExportModelSettings.info = exportModelSettingsSerialize;
             DisplayOptionsWindow = true;
-            ConvertToPrefabSettings = ScriptableObject.CreateInstance (typeof(ConvertToPrefabSettings)) as ConvertToPrefabSettings;
-            convertToPrefabSettingsSerialize = ConvertToPrefabSettings.info;
+            convertToPrefabSettingsSerialize = new ConvertToPrefabSettingsSerialize();
+            ConvertToPrefabSettings.info = convertToPrefabSettingsSerialize;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


com.unity.formats.fbx/Editor/FbxExportSettings.cs
```diff
@@ -863,8 +868,8 @@ internal void LoadDefaults()
             prefabSavePaths = new List<string>(){ kDefaultSavePath };
             fbxSavePaths = new List<string> (){ kDefaultSavePath };
             integrationSavePath = DefaultIntegrationSavePath;
-            dccOptionPaths = null;
-            dccOptionNames = null;
+            dccOptionPaths = new List<string>();
+            dccOptionNames = new List<string>();
```
natalia-wh, line 872: Changed because tests were failing. `dccOptionPaths` and `dccOptionNames` are initialized to string lists, so resetting made sense to be string lists  unless they need to be null for another reason?

com.unity.formats.fbx/Editor/FbxExportSettings.cs
```diff
@@ -1745,6 +1750,24 @@ internal void Save()
             convertToPrefabSettingsSerialize = ConvertToPrefabSettings.info;
             this.SaveToFile ();
         }
+
+        // Called on creation and whenever the reset button is clicked
+        internal void Reset()
+        {
+            // create presets if none exist
+            if (!m_defaultModelExportSettings) {
+                m_defaultModelExportSettings = new Preset(ExportModelSettings);
```
vkovec, line 1759: I'm not sure I understand the use of these presets, is it to reset the ExportModelSettings to default? would it work to set `exportModelSettingsSerialize` to null?

vkovec (commented):
can you please add a test that calls Reset() to the FbxExporterSettingsTests?

com.unity.formats.fbx/Editor/FbxExportSettings.cs
natalia-wh: Yeah, the presets are for resetting `ExportModelSettings` and `ConvertToPrefabSettings` to default. 

Reset is called both on creation and when hitting the reset button, so theyre saved on creation and re-copied over whenever reset is clicked.

com.unity.formats.fbx/Editor/FbxExportSettings.cs
```diff
@@ -476,6 +477,10 @@ public enum ObjectPosition { LocalCentered = 0, WorldAbsolute = 1, Reset = 2 /*
 
         public enum LODExportType { All = 0, Highest = 1, Lowest = 2 }
 
+        // presets for editor settings
+        private Preset m_defaultModelExportSettings;
```
vkovec, line 481: please remove these if no longer used

vkovec (approved):

PR #596, opened 2021-06-07
FBX 92 Fix prefab rename on undo

natalia-wh:
Fix prefab staying renamed after undoing prefab creation

### DIFF
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -2790,10 +2790,10 @@ namespace UnityEditor.Formats.Fbx.Exporter
             string fbxName = unityGo.name;
             if (ExportOptions.UseMayaCompatibleNames)
             {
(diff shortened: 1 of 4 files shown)


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -2790,6 +2790,7 @@ private FbxNode CreateFbxNode(GameObject unityGo, FbxScene fbxScene)
             string fbxName = unityGo.name;
             if (ExportOptions.UseMayaCompatibleNames)
             {
+                Undo.RecordObject(unityGo, "rename " + fbxName);
```
vkovec, line 2793: could this be placed in the `if (ExportOptions.AllowSceneModification)` a couple of lines below?
The scene will only be modified if the option is true.

vkovec (commented):
can you please add a unit test that converts a hierarchy with name changes then calls undo and checks the result?

vkovec (approved):

PR #599, opened 2021-06-09
FBX-107 Fix warning about ExportSettings

natalia-wh:
Fix warning about ExportSettings class not being in a file called `ExportSettings.cs`
Rename `FbxExportSettings.cs` to `ExportSettings.cs`


com.unity.formats.fbx/Editor/FbxExportSettings.cs.meta
```diff
@@ -1,11 +0,0 @@
-fileFormatVersion: 2
-guid: 03d96513d037ea843ab436321bab4c33
-MonoImporter:
```
vkovec, line 3: can you please rename the meta file as well? Could give warnings about missing meta file if we do not include it.

com.unity.formats.fbx/Editor/ExportSettings.cs.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: 03d96513d037ea843ab436321bab4c33
+guid: f87a2d71b287d4b579dbf0858cda24d9
```
vkovec, line 2: would it be possible to keep the same guid? we used to have issues with changing them. Not sure if it would still be a problem, but just to be safe.

vkovec (approved):

PR #613, opened 2021-11-18
FBX-220 handle exporting more than 4 bone weights

vkovec:


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -1199,42 +1197,68 @@ private bool ExportSkeleton (SkinnedMeshRenderer skinnedMesh, FbxScene fbxScene,
         /// <summary>
         /// set vertex weights in cluster
         /// </summary>
-        private void SetVertexWeights (MeshInfo meshInfo, Dictionary<int, FbxCluster> boneCluster)
+        private void SetVertexWeights (MeshInfo meshInfo, Dictionary<int, FbxCluster> boneIndexToCluster)
         {
-            HashSet<int> visitedVertices = new HashSet<int> ();
+            var mesh = meshInfo.mesh;
```
vkovec, line 1202: changes are mostly taken from the documentation: https://docs.unity3d.com/ScriptReference/Mesh.GetAllBoneWeights.html

com.unity.formats.fbx/Tests/Models/multiBoneInfluence.fbx
```diff
@@ -0,0 +1,947 @@
+; FBX 7.7.0 project file
+; ----------------------------------------------------
```
vkovec, line 2: Just a cube with 8 bones.

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
```diff
@@ -712,53 +717,72 @@ public class SkinnedMeshTestDataClass
...
+            // Get the number of bone weights per vertex
+            var origBonesPerVertex = origMesh.GetBonesPerVertex();
+            var exportedBonesPerVertex = exportedMesh.GetBonesPerVertex();
+
+            // Get all the bone weights, in vertex index order
+            var origBoneWeights1 = origMesh.GetAllBoneWeights();
+            var exportedBoneWeights1 = exportedMesh.GetAllBoneWeights();
+
+            // get the vertex index to index in bone weights mapping (for easier access)
+            var origVertToBoneWeightIndex = new Dictionary<int, int>();
```
benoithudson, line 734: this and `exportedVertToBoneWeightIndex` could be arrays I think?

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
```diff
@@ -712,53 +717,72 @@ public class SkinnedMeshTestDataClass
...
 
-                        var indexMsg = "Bone index {0} doesn't match";
-                        var nameMsg = "bone names don't match";
+            var comparisonCount = 0;
+            var minVertCount = Mathf.Min(origVerts.Length, expVerts.Length);
+            for (int origVertIndex = 0; origVertIndex < minVertCount; origVertIndex++)
+            {
+                for (int expVertIndex = 0; expVertIndex < minVertCount; expVertIndex++)
+                {
+                    if (origVerts[origVertIndex] == expVerts[expVertIndex])
```
benoithudson, line 756: I think this bit is wrong.
(1) It's possible to have two vertices in the same location in a mesh; this will only find the first one and assign all the bones to it.
(2) It's quadratic time. In a 100k or 1M vertex mesh, that might take a long time.

Why would the indices change between the original mesh and the export mesh?

benoithudson: Oh wait I'm confused, this is testing not exporting.

I think we should test that the indices don't change; they shouldn't (some things depend on vertex indexes not changing).

benoithudson (approved):
all my comments are in the tests; I  think you can ignore them

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
vkovec: yep

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
vkovec: I think this is because there the vertex count did not always match before/after export. We were doing the same before as well.

PR #617, opened 2022-04-01
CI: update for 4.0 + add nightlies

mfe:


mfe:
Quoting a wise person:
> 2021.1 is no longer supported, so 4.0 is effectively also unsupported

So closing this PR !

PR #619, opened 2022-04-02
Feat: Add support for texture offset/scale/wrapMode

hybridherbst:
This PR adds support for texture offset, scale, and wrap mode.
Since FBX seems to not have a concept for "same texture used in different ways", it looks like textures will be duplicated on export.

I added a warning message if the wrap mode of the texture can't be exported (FBX only supports Clamp and Repeat), not sure if that's in line with other areas here (seems the general rule of thumb is that the FBX Exporter will silently ignore errors, which isn't great).

### DIFF
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -175,9 +175,9 @@ namespace UnityEditor.Formats.Fbx.Exporter
         Dictionary<int, FbxSurfaceMaterial> MaterialMap = new Dictionary<int, FbxSurfaceMaterial> ();
 
         /// <summary>
-        /// Map texture filename name to FBX texture object
+        /// Map texture properties to FBX texture object
         /// </summary>
-        Dictionary<string, FbxTexture> TextureMap = new Dictionary<string, FbxTexture> ();
+        Dictionary<(Texture unityTexture, Vector2 offset, Vector2 scale, TextureWrapMode wrapModeU, TextureWrapMode wrapModeV), FbxTexture> TextureMap = new Dictionary<(Texture unityTexture, Vector2 offset, Vector2 scale, TextureWrapMode wrapModeU, TextureWrapMode wrapModeV), FbxTexture>();
 
         /// <summary>
         /// Map a Unity mesh to an fbx node (for preserving instances) 
@@ -635,20 +635,41 @@ namespace UnityEditor.Formats.Fbx.Exporter
                 return false;
             }
 
+            var offset = unityMaterial.GetTextureOffset(unityPropName);
+            var scale = unityMaterial.GetTextureScale(unityPropName);
+            var wrapModeU = unityTexture.wrapModeU;
+            var wrapModeV = unityTexture.wrapModeV;
+
+            var tuple = (unityTexture, offset, scale, wrapModeU, wrapModeV);
+            
             // Find or create an fbx texture and link it up to the fbx material.
-            if (!TextureMap.ContainsKey (textureSourceFullPath)) {
+            if (!TextureMap.ContainsKey (tuple)) {
                 var textureName = GetUniqueTextureName(fbxPropName + "_Texture");
                 var fbxTexture = FbxFileTexture.Create (fbxMaterial, textureName);
                 fbxTexture.SetFileName (textureSourceFullPath);
                 fbxTexture.SetTextureUse (FbxTexture.ETextureUse.eStandard);
                 fbxTexture.SetMappingType (FbxTexture.EMappingType.eUV);
-                TextureMap.Add (textureSourceFullPath, fbxTexture);
+                fbxTexture.SetScale(scale.x, scale.y);
+                fbxTexture.SetTranslation(offset.x, offset.y);
+                fbxTexture.SetWrapMode(GetWrapModeFromUnityWrapMode(wrapModeU), GetWrapModeFromUnityWrapMode(wrapModeV));
+                TextureMap.Add (tuple, fbxTexture);
             }
-            TextureMap [textureSourceFullPath].ConnectDstProperty (fbxMaterialProperty);
+            TextureMap [tuple].ConnectDstProperty (fbxMaterialProperty);
 
             return true;
         }
 
+        private FbxTexture.EWrapMode GetWrapModeFromUnityWrapMode(TextureWrapMode wrapMode)
+        {
+            switch (wrapMode)
+            {
+                case TextureWrapMode.Clamp: return FbxTexture.EWrapMode.eClamp;
+                case TextureWrapMode.Repeat: return FbxTexture.EWrapMode.eRepeat;
+                default:
+                    Debug.LogWarning("Texture uses wrap mode " + wrapMode + " which is not supported by FBX. Will be exported as \"Repeat\".");
+                    return FbxTexture.EWrapMode.eRepeat;
+            }
+        }
         /// <summary>
         /// Get the color of a material, or grey if we can't find it.
         /// </summary>


vkovec:
awesome! for the tests, would it be possible to add a unit test in ModelExporterTests.cs exporting and checking the reloaded material? Here is an example test: https://github.com/Unity-Technologies/com.unity.formats.fbx/blob/master/com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs#L1136

The assets needed for the test can also be added to a new folder (e.g. Textures or Materials) one level up: https://github.com/Unity-Technologies/com.unity.formats.fbx/tree/master/com.unity.formats.fbx/Tests

hybridherbst:
Added a test and assets. The test assets look like this when applied to a default cube:
image

Note that the wrap mode test isn't actually correct yet - to test wrap modes (or any actual roundtrip behaviour), one would need to
- export with embedded textures
- import
- unpack textures
- check wrap mode

but I haven't added that, think that might be out of scope of this PR.

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
```diff
@@ -1227,5 +1228,72 @@ public void TestMissingMeshExport()
...
+            var cubeRepeat = GameObject.CreatePrimitive(PrimitiveType.Cube);
+            cubeRepeat.name = "Repeat";
+            cubeRepeat.GetComponent<MeshRenderer>().sharedMaterial = repeatMat;
+
+            var settings = new ExportModelSettingsSerialize();
+            settings.SetEmbedTextures(true);
+            ModelExporter.ExportObjects(filename, new Object[] { cubeClamp, cubeRepeat }, settings);
+
+            var modelImporter = AssetImporter.GetAtPath(filename) as ModelImporter;
+            modelImporter.ExtractTextures(tempUnpackPath);
```
vkovec, line 1255: To get a random path in assets you can call `GetRandomFileNamePath(extName: "");`.
The base class will automatically delete it on teardown.

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
```diff
@@ -1227,5 +1228,72 @@ public void TestMissingMeshExport()
             ModelExporter.ExportObject(filename, cube);
             Assert.IsNotNull(filename);
         }
+
+        [Test]
+        public void TestMaterialScaleAndOffset() 
+        {
+            string matClampGuid = "e6598d8bd228e5940988877e9eddd594";
+            string matRepeatGuid = "f07bd71d1d87a7b41acfc483bf06be3f";
```
vkovec, line 1236: this var doesn't seem to be used, I think there is a typo on line 1245 which uses the clamp guid instead of this one

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
```diff
@@ -1227,5 +1228,72 @@ public void TestMissingMeshExport()
...
+            // Ensure we're actually checking on the newly imported textures, not on the ones used for export
+            Assert.NotNull(importedClampMat.mainTexture);
+            Assert.NotNull(importedRepeatMat.mainTexture);
+            
+            // (Case 1424290) This fails - seems there's a bug with modelImporter.ExtractTextures when used inside a test
+            // Manually exporting the FBX and extracting textures via Inspector properly produces separate textures.
+            // Assert.AreNotSame(importedClampMat.mainTexture, importedRepeatMat.mainTexture);
+            
+            // Seems these also randomly fail? Sometimes work, sometimes not...
+            // Assert.AreNotSame(clampMat.mainTexture, importedClampMat.mainTexture);
```
vkovec, line 1271: the texture extraction in the test seems like it might not update the material to point to the extracted textures. If I run the test without deleting the generated asset, the materials seem to point to the original, not the embedded ones. If I export and extract the textures manually, it points to the extracted textures.

vkovec: oh sorry this is the bug commented right above, I misunderstood. I'm not sure if waiting a few frames would solve the issue, I think there is an extra remap done after ExtractTextures is called when you click the button in the UI.

Extract textures call: https://github.com/Unity-Technologies/UnityCsReference/blob/master/Modules/AssetPipelineEditor/ImportSettings/ModelImporterMaterialEditor.cs#L299

Remap call: https://github.com/Unity-Technologies/UnityCsReference/blob/master/Modules/AssetPipelineEditor/ImportSettings/ModelImporterMaterialEditor.cs#L326

I'm not sure if the remap is easy to replicate here, or maybe could find the matching textures in the folder they were extracted to and load the texture asset directly to check the values. If it is difficult to do, then in my opinion since there is the other bug with the wrap mode not importing correctly, this could also potentially be fixed when that issue is resolved.

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
hybridherbst: Nice, that removes the extra Teardown method again.

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
hybridherbst: Fixed, good catch

com.unity.formats.fbx/Tests/FbxTests/ModelExporterTest.cs
hybridherbst: Turns out it was easy to replicate - thanks for pointing me at the source there, wasn't aware that Extract exactly only does that and not the remapping part. One bug down! Only the import wrap mode issue remains, but that code can just be uncommented once that's at some point fixed in core.

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -635,20 +635,41 @@ internal static FbxVector4 ConvertToFbxVector4(Vector3 leftHandedVector, float u
...
         }
 
+        private FbxTexture.EWrapMode GetWrapModeFromUnityWrapMode(TextureWrapMode wrapMode)
+        {
+            switch (wrapMode)
+            {
+                case TextureWrapMode.Clamp: return FbxTexture.EWrapMode.eClamp;
+                case TextureWrapMode.Repeat: return FbxTexture.EWrapMode.eRepeat;
+                default:
+                    Debug.LogWarning("Texture uses wrap mode " + wrapMode + " which is not supported by FBX. Will be exported as \"Repeat\".");
```
benoithudson, line 669: I'm not convinced we should issue this warning. When it comes to materials we don't claim to transfer losslessly.

If we do really want to issue a warning I'd recommend at least including a name that the user can search for to go check on maybe they want to fix it and export again. And maybe an object for context.

benoithudson (changes requested):
Ensure either there's no warning, or add some information so the warning can be useful to the user.

Optional minor performance note as well, up to you to do or not.

vkovec (approved):

com.unity.formats.fbx/Editor/FbxExporter.cs
vkovec: will fix in another PR

hybridherbst:
Hey @vkovec, I'm a bit confused by GitHubs UI here - are there still requested changes? I thought I had resolved all.

vkovec:
@hybridherbst sorry for the confusion, I am trying to merge the PR. There are a couple of minor change requests from Benoit, I was going to fix those after merging.

hybridherbst:
Thank you!

PR #620, opened 2022-04-08
Update master from release 4.1

vkovec:


com.unity.formats.fbx/package.json
```diff
@@ -16,4 +16,4 @@
     "maya",
     "max"
   ]
-}
+}
```
benoithudson, line 19: can you drop this change? different text editors fight over whether there's a newline at the end

benoithudson (approved):
if you can easily drop the package.json change that's ideal

com.unity.formats.fbx/package.json
vkovec: discussed offline to leave this as it comes from a commit from the merge

PR #627, opened 2022-04-22
fix directory not deleted at end of test

vkovec:


com.unity.formats.fbx/Tests/FbxApiTests/SampleCodeTest.cs
```diff
@@ -11,35 +11,18 @@ namespace FbxExporter.UnitTests
...
-            Assert.That(m_tempFolder, Does.Not.Exist);
-            Directory.CreateDirectory(m_tempFolder);
-        }
-
-        [TearDown]
-        public void Term()
-        {
-            if (Directory.Exists(m_tempFolder))
-            {
-                Directory.Delete(m_tempFolder, recursive: true);
```
vkovec, line 31: this was not deleting the directory properly, and so if you run the test again it would fail on init as the directory already exists. Reusing the code from the test base for creating/deleting directories

benoithudson (approved):

com.unity.formats.fbx/Tests/FbxApiTests/SampleCodeTest.cs
```diff
@@ -54,7 +35,9 @@ public void TestExportGameObjectsSample()
             var exportPath = Path.Combine(Application.dataPath, filename);
             Assert.That(exportPath, Does.Exist);
 
-            Object[] loaded = AssetDatabase.LoadAllAssetsAtPath("Assets/" + filename);
+            var assetPath = "Assets/" + filename;
```
vkovec, line 38: had to keep this test mostly the same to work with the sample function

PR #630, opened 2022-05-06
run nightlies on release/4.2

vkovec:


benoithudson (approved):
If I were in charge of the cluster I'd prefer if my users only ran nightlies against trunk (on every platform, sure). But I'm not in charge.

mfe:
I think what we want is for each release branch (release/4.1, release/4.2...) to have nightlies that run every day that tests the versions we are supposed to support for this specific versions: 
- the version we released against
- LTS versions we want to support
- trunk (important to monitor when trunk breaks us)

So for release/4.1:
- 2022.1 
- probably 2019.4, 2020.3, 2021.3
- I would skip trunk here

for release/4.2:
- 2022.2
- probably 2019.4, 2020.3, 2021.3
- trunk

vkovec:
will create a new branch/PR for this

PR #632, opened 2022-05-17
Nojira/update nightly test editor versions

vkovec:
updating the nightly editor versions based on this comment: https://github.com/Unity-Technologies/com.unity.formats.fbx/pull/630#issuecomment-1124703236

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -1,7 +1,27 @@
-test_trigger_editors:
+release_4_2_nightly_trigger_editors:
   - version: 2019.4
   - version: 2020.3
   - version: 2021.3
+  - version: 2022.2
+  - version: trunk
+
+release_4_1_nightly_trigger_editors:
+  - version: 2019.4
+  - version: 2020.3
+  - version: 2021.3
+  - version: 2022.1
+  
+master_nightly_trigger_editors:
+  - version: 2019.4
+  - version: 2020.3
+  - version: 2021.3
+  - version: 2022.2
+  - version: trunk
+
+test_trigger_editors:
+  - version: 2019.4
+  - version: 2020.3
+  - version: 2021.2
   - version: 2022.1
   - version: trunk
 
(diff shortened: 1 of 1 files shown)


.yamato/upm-ci.yml
```diff
@@ -1,3 +1,23 @@
+release_4_2_nightly_trigger_editors:
```
benoithudson, line 1: Can you add a comment (line that starts with '#') to describe what we should do when we release another version, when to update these, etc?

In particular I'm wondering why we have a mix of 2022.2 and 2022.1

.yamato/upm-ci.yml
```diff
@@ -104,19 +124,46 @@ test_trigger:
 {% endfor %}
 {% endfor %}
 
-nightly_test_trigger:
+release_4_2_nightly_test_trigger:
```
benoithudson, line 127: is it possible to do 
`{% for release in nightly_tested_releases %}` here and

 `     {% for editor in {{release}}_nightly_trigger_editors %}`
later (currently line 136)?

The answer used to be no but they were looking into fixing that. If we can do indirect references like that we could save a lot of repetition.

.yamato/upm-ci.yml
```diff
@@ -104,19 +124,46 @@ test_trigger:
...
+  dependencies:
+    - .yamato/upm-ci.yml#pack
+      {% for editor in release_4_2_nightly_trigger_editors %}
+      {% for platform in platforms %}
+    - .yamato/upm-ci.yml#test_{{platform.name}}_{{editor.version}}
+      {% endfor %}
+      {% endfor %}
+      
+release_4_1_nightly_test_trigger:
+  name: Nightly tests Trigger
```
benoithudson, line 143: include the release early on in the name so it's easy to spot on yamato, e.g. "nightly release/4.1 tests trigger"

.yamato/upm-ci.yml
vkovec: for the mix, I followed the versions from Marie's comment here: https://github.com/Unity-Technologies/com.unity.formats.fbx/pull/630#issuecomment-1124703236

I think it is the latest version that was available when that version was released

benoithudson (approved):

PR #640, opened 2022-07-04
FBX-95 change property channel pairs to list

vkovec:


com.unity.formats.fbx/Editor/FbxPropertyChannelPair.cs
```diff
@@ -31,122 +31,150 @@ public UnityPropertyChannelPair(string p, string c)
...
+            public PropertyChannelMap(List<(string, string)> propertyMap, List<(string, string)> channelMap)
             {
                 MapUnityPropToFbxProp = propertyMap;
                 MapUnityChannelToFbxChannel = channelMap;
             }
+
+            private string GetFbxValue(string uniValue, List<(string,string)> list)
+            {
+                var index = list.FindIndex(x => x.Item1 == uniValue);
+                return index < 0 ? null : list[index].Item2;
```
benoithudson, line 51: does `return list.FirstOrDefault(x => x.Item1 == uniValue).Item2` work? 

If not then this is good too.

benoithudson (approved):

com.unity.formats.fbx/Editor/FbxPropertyChannelPair.cs
vkovec: there isn't a first or default, but I think Find gives the same result

PR #643, opened 2022-07-06
FBX-28 expose export options

vkovec:


com.unity.formats.fbx/Editor/ExportSettings.cs
```diff
@@ -10,6 +10,43 @@
...
+    /// <summary>
+    /// Enum specifying the FBX export format options.
+    /// </summary>
+    public enum ExportFormat { ASCII = 0, Binary = 1 }
+
+    /// <summary>
+    /// Enum specifying the type of data to include in the export
+    /// (Model only, animation only, or model and animation).
+    /// </summary>
+    public enum Include { Model = 0, Anim = 1, ModelAndAnim = 2 }
```
benoithudson, line 23: if we have `Model=1` `Anim=2` and `ModelAndAnim=3` then you can do `Include.Anim | Include.Model`

benoithudson (approved):

com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -4527,14 +4527,43 @@ private static void OnExport ()
...
-            return ExportObjects(filePath, new Object[] { root }, exportOptions);
+            return ExportObjects(filePath, new Object[] { singleObject }, exportOptions, exportData: null);
+        }
+
+        /// <summary>
+        /// Exports an array of Unity GameObjects to an FBX file,
+        /// with the specified export settings.
+        /// </summary>
+        /// <returns>
+        /// The FBX file path if successful; otherwise returns null.
```
sebastienduverne, line 4555:
```suggestion
        /// The FBX file path if successful; otherwise null.
```

sebastienduverne (commented):
Multiple suggestions for API docs.
**Note:** Use the "Add suggestion to batch" button if you want to apply several suggestions in batch through a single commit.

com.unity.formats.fbx/Editor/Unity.Formats.Fbx.Editor.api
```diff
@@ -0,0 +1,125 @@
...
+        public void SetEmbedTextures(bool embedTextures);
+        public void SetExportUnredererd(bool exportUnrendered);
+        public void SetKeepInstances(bool keepInstances);
+        public void SetLODExportType(UnityEditor.Formats.Fbx.Exporter.LODExportType lodLevel);
+        public void SetModelAnimIncludeOption(UnityEditor.Formats.Fbx.Exporter.Include include);
+        public void SetObjectPosition(UnityEditor.Formats.Fbx.Exporter.ObjectPosition objectPosition);
+        public void SetPreserveImportSettings(bool preserveImportSettings);
+    }
+
+    public abstract class ExportOptionsSettingsSerializeBase : UnityEditor.Formats.Fbx.Exporter.IExportOptions
```
clusty, line 50: This seems to add a lot of noise for the public API.
Would it make sense to make certain things only read and not write and have 2 factories for the 2 settings types ?

com.unity.formats.fbx/Editor/Unity.Formats.Fbx.Editor.api
```diff
@@ -0,0 +1,125 @@
...
+        public virtual bool Equals(object e);
+        public virtual int GetHashCode();
+        public void SetAnimatedSkinnedMesh(bool animatedSkinnedMesh);
+        public void SetAnimationDest(UnityEngine.Transform dest);
+        public void SetAnimationSource(UnityEngine.Transform source);
+        public void SetExportFormat(UnityEditor.Formats.Fbx.Exporter.ExportFormat format);
+        public void SetUseMayaCompatibleNames(bool useMayaCompNames);
+    }
+
+    public interface IExportOptions
```
clusty, line 75: can you make the interface internal and have a "hidden implementation" ?

com.unity.formats.fbx/Editor/Unity.Formats.Fbx.Editor.api
```diff
@@ -0,0 +1,125 @@
...
+        public abstract UnityEditor.Formats.Fbx.Exporter.ObjectPosition ObjectPosition { get; }
+        public virtual bool PreserveImportSettings { get; }
+        public virtual bool UseMayaCompatibleNames { get; }
+        protected ExportOptionsSettingsSerializeBase() {}
+        public virtual bool Equals(object e);
+        public virtual int GetHashCode();
+        public void SetAnimatedSkinnedMesh(bool animatedSkinnedMesh);
+        public void SetAnimationDest(UnityEngine.Transform dest);
+        public void SetAnimationSource(UnityEngine.Transform source);
+        public void SetExportFormat(UnityEditor.Formats.Fbx.Exporter.ExportFormat format);
```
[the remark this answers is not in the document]
vkovec: I don't remember why we used a mix of properties and functions. Will check with Benoit but I don't see why I couldn't change it to use get functions.

com.unity.formats.fbx/Editor/Unity.Formats.Fbx.Editor.api
vkovec: possibly, I'll see what I can do

com.unity.formats.fbx/Editor/Unity.Formats.Fbx.Editor.api
```diff
@@ -0,0 +1,125 @@
+// This file is generated. Do not modify by hand.
+// XML documentation file not found. To check if public methods have XML comments,
+// make sure the XML doc file is present and located next to the scraped dll
+namespace UnityEditor.Formats.Fbx.Exporter
+{
+    public static class ConvertToNestedPrefab
+    {
+        public static UnityEngine.GameObject Convert(UnityEngine.GameObject toConvert, string fbxDirectoryFullPath = default(string), string fbxFullPath = default(string), string prefabDirectoryFullPath = default(string), string prefabFullPath = default(string), UnityEditor.Formats.Fbx.Exporter.ConvertToPrefabSettingsSerialize exportOptions = default(UnityEditor.Formats.Fbx.Exporter.ConvertToPrefabSettingsSerialize));
```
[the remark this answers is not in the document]
benoithudson: Export doesn't change the scene (or if it does there's a bug we should fix someday).

Convert does change the scene. We're trying to emulate the semantics that users see when they drag a gameobject from the hierarchy to the project folder. In that case, what you see in the scene is that the gameobject you dragged becomes an instance of the new prefab, and all references in the scene to anything in that hierarchy stays connected to it.

com.unity.formats.fbx/Editor/Unity.Formats.Fbx.Editor.api
```diff
@@ -0,0 +1,125 @@
...
+
+    public enum LODExportType
+    {
+        public const UnityEditor.Formats.Fbx.Exporter.LODExportType All = 0;
+        public const UnityEditor.Formats.Fbx.Exporter.LODExportType Highest = 1;
+        public const UnityEditor.Formats.Fbx.Exporter.LODExportType Lowest = 2;
+        public int value__;
+    }
+
+    public sealed class ModelExporter : System.IDisposable
```
benoithudson, line 108: Is this the time when we can finally get rid of IDisposable here? We aren't actually disposable, it's all static.

com.unity.formats.fbx/Editor/Unity.Formats.Fbx.Editor.api
benoithudson: @clusty : if I understand, you're proposing the public API be something like:
* Convert(ConvertSettings) with a public class ConvertSettings
* Export(ExportSettings) with a public class ExportSettings
The two classes are similar but have no relationship.

Then under the hood we can convert each settings class to the current overengineered hierarchy that we currently have.

sebastienduverne (approved):

com.unity.formats.fbx/Editor/Unity.Formats.Fbx.Editor.api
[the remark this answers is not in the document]
vkovec: >If you set this paths to things outside of "Assets" folder does it work ? (I see those paths are sent to prefab functions)

it will throw an exception in this case, I have added unit tests for it.

clusty (approved):

PR #649, opened 2022-07-12
FBX-307 optimizations to convert to prefab variant

vkovec:


com.unity.formats.fbx/Editor/ConvertToNestedPrefab.cs
```diff
@@ -221,7 +221,7 @@ internal static bool SetContainsOnlyPrefabAssets(Object[] toConvert)
         /// <param name="origObj">original object</param>
         /// <param name="newObj">new object</param>
         /// <param name="toConvertRoot">root of the hierarchy being converted</param>
-        internal static void FixSceneReferenceToObject(Object sceneObj, Object origObj, Object newObj, GameObject toConvertRoot)
+        internal static void FixSceneReferenceToObject(Object sceneObj, GameObject toConvertRoot, Dictionary<Object, Object> fixSceneRefsMap)
```
benoithudson, line 224: You'll need to update the comment.

com.unity.formats.fbx/Editor/ConvertToNestedPrefab.cs
```diff
@@ -526,6 +535,56 @@ internal static List<GameObject> GetSceneReferencesToObject(Object obj)
...
+                var info = new SourceObjectInfo();
+#if UNITY_2021_2_OR_NEWER
+                if (!isPreviewScene)
+                {
+                    var instanceID = t.gameObject.GetInstanceID();
+                    info.sceneObjectsWithReference = GetSceneReferences(instanceID);
+                }
+#endif // UNITY_2021_2_OR_NEWER
+
+                s_nameToInfo[t.name] = info;
```
benoithudson, line 579: How do you set sceneObjectsWithReference in 2020.3 ?

com.unity.formats.fbx/Editor/ConvertToNestedPrefab.cs
```diff
@@ -526,6 +535,56 @@ internal static List<GameObject> GetSceneReferencesToObject(Object obj)
...
+            while (q.Count > 0)
+            {
+                var t = q.Dequeue();
+
+                var info = new SourceObjectInfo();
+#if UNITY_2021_2_OR_NEWER
+                if (!isPreviewScene)
+                {
+                    var instanceID = t.gameObject.GetInstanceID();
+                    info.sceneObjectsWithReference = GetSceneReferences(instanceID);
```
benoithudson, line 575: This means: for all objects t that are being converted, loop over the entire scene to get the references to t.

Better than before when it was for all *components* loop over the entire scene but still quadratic.

If this is fast enough then we can leave it there but we could theoretically do better.

com.unity.formats.fbx/Editor/ConvertToNestedPrefab.cs
benoithudson: Also: I assume we renamed objects in the hierarchy before we get to this point, right?

com.unity.formats.fbx/Editor/ConvertToNestedPrefab.cs
```diff
@@ -786,8 +847,38 @@ internal static void UpdateFromSourceRecursive(GameObject dest, GameObject sourc
...
+                    foreach (var item in items)
+                    {
+                        FixSceneReferenceToObject(item, source, fixSceneRefsMap);
+                    }
+                }
+#else // UNITY_2021_2_OR_NEWER
+                foreach (var obj in fixSceneRefsMap.Keys)
+                {
+                    // find and fix the references in the scene
+                    var sceneObjs = GetSceneReferencesToObject(obj);
```
benoithudson, line 873: can you do this earlier and stuff it into info.sceneObjectsWithReference, then remove the #if here?

com.unity.formats.fbx/Editor/ConvertToNestedPrefab.cs
```diff
@@ -873,13 +953,14 @@ internal static void UpdateFromSourceRecursive(GameObject dest, GameObject sourc
...
         /// <param name="nameMap"></param>
-        internal static void CopySerializedProperty(SerializedObject serializedObject, SerializedProperty fromProperty, Dictionary<string, GameObject> nameMap)
+        internal static void CopySerializedProperty(SerializedObject serializedObject, SerializedProperty fromProperty)
         {
             var toProperty = serializedObject.FindProperty(fromProperty.propertyPath);
 
-            GameObject value;
-            if (nameMap.TryGetValue(fromProperty.objectReferenceValue.name, out value))
+            SourceObjectInfo info;
+            if (s_nameToInfo.TryGetValue(fromProperty.objectReferenceValue.name, out info))
```
benoithudson, line 961: You could do `out var info` here.

com.unity.formats.fbx/Editor/ConvertToNestedPrefab.cs
```diff
@@ -526,6 +535,56 @@ internal static List<GameObject> GetSceneReferencesToObject(Object obj)
...
+
+        internal static void GatherSceneHierarchy(GameObject hierarchyRoot)
+        {
+            s_nameToInfo.Clear();
+
+#if UNITY_2021_2_OR_NEWER
+            var isPreviewScene = SceneManagement.EditorSceneManager.IsPreviewSceneObject(hierarchyRoot);
+#endif // UNITY_2021_2_OR_NEWER
+
+            var q = new Queue<Transform>();
```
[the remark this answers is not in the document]
vkovec: I don't think it matters, I don't remember why we used a queue before

com.unity.formats.fbx/Editor/ConvertToNestedPrefab.cs
vkovec: I think I could make it so that it does one huge query but then it will be more difficult to figure out which component reference needs to be fixed in which result.

com.unity.formats.fbx/Editor/ConvertToNestedPrefab.cs
vkovec: I don't because in 2020.3 searching for the object does not return results of objects that reference its components

benoithudson: OK now I get it. Can you document that somehow, that you can search by object in recent Unity but in older Unity you need to search by component?

benoithudson (approved):

PR #656, opened 2022-07-29
Nojira/fix fbxrecorder with 4.0.0 recorder

vkovec:
Fix an error when recording from a recorder track (end recording is called twice, need to check if the recording has already ended to not try to export twice).

Also fix the animation compression setting not being used.

### DIFF
--- a/com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorder.cs
+++ b/com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorder.cs
@@ -12,6 +12,60 @@ namespace UnityEditor.Formats.Fbx.Exporter
         {
         }
 
+        private void EndRecordingInternal(RecordingSession session)
+        {
+            var settings = (FbxRecorderSettings)session.settings;
+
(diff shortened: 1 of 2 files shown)


com.unity.formats.fbx/Editor/Sources/Recorders/FbxRecorder/FbxRecorder.cs
```diff
@@ -21,54 +19,75 @@ protected override void EndRecording(RecordingSession session)
                 throw new System.ArgumentNullException("session");
             }
 
-            var settings = (FbxRecorderSettings)session.settings;
-
-            foreach (var input in m_Inputs)
+            if (Recording)
```
benoithudson, line 22: This is error-prone: it's very easy for someone to stick a `return` in the middle of this big block and then the `            base.EndRecording(session);` doesn't happen.

What you can do is make it be:
```
if (Recording)
{
   ActuallyEndRecording(session);
}
base.EndRecording(session);
```
and have all that code in this block be in a separate function.

(Side benefit: it means you avoid reindenting it)

benoithudson (approved):

PR #652, opened 2022-07-23
FbxExporter.cs improvement

smitdylan2001:
Use count and length properties instead of LINQ call.
Array/List can not be under 0.
Use TryGetComponent to reduce memory allocation when no component is found.
GetComponent Mesh will always return null. Get MeshFilter.mesh instead.
Simplify string comparison.

PR #657, opened 2022-07-29
FbxExporter.cs Improvements

smitdylan2001:
New PR to fix issues with following PR: https://github.com/Unity-Technologies/com.unity.formats.fbx/pull/652

Contents:
Use count and length properties instead of LINQ call.
Array/List can not be under 0, so array size will be checked as == 0 instead of <= 0.
Use TryGetComponent to reduce memory allocation when no component is found.
GetComponent Mesh will always return null. Get MeshFilter.mesh instead.
Simplify string comparison using IndexOf, which does not create garbage.
Cache animation controller layers to avoid an extra internal call

### DIFF
--- a/com.unity.formats.fbx/Editor/FbxExporter.cs
+++ b/com.unity.formats.fbx/Editor/FbxExporter.cs
@@ -1,4 +1,3 @@
-
 using System.IO;
 using System.Collections.Generic;
 using UnityEngine;
@@ -1039,7 +1038,6 @@ namespace UnityEditor.Formats.Fbx.Exporter
                 return false;
             }
 
-            
             if (!unityGo.TryGetComponent<SkinnedMeshRenderer>(out var unitySkin))
             {
                 return false;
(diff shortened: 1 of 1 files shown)


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -771,9 +771,10 @@ internal bool ExportMaterial(Material unityMaterial, FbxScene fbxScene, FbxNode
 
             // We'll export either Phong or Lambert. Phong if it calls
             // itself specular, Lambert otherwise.
+            System.StringComparison strignComparison = System.StringComparison.OrdinalIgnoreCase;
```
vkovec, line 774: small typo on the variable name

vkovec (approved):

com.unity.formats.fbx/Editor/FbxExporter.cs
smitdylan2001: Fixed

benoithudson (approved):
Thanks, looks great!

PR #667, opened 2022-09-16
FBX-356 fix package settings broken in Ubuntu

vkovec:


com.unity.formats.fbx/Editor/ExportSettings.cs
```diff
@@ -1183,17 +1184,19 @@ private static void FindDCCInstalls()
             var dccOptionNames = instance.dccOptionNames;
             var dccOptionPaths = instance.dccOptionPaths;
 
+            var vendorLocations = DCCVendorLocations;
```
vkovec, line 1187: small optimization to not call the getter multiple times

com.unity.formats.fbx/Editor/ExportSettings.cs
```diff
@@ -762,7 +762,8 @@ private static HashSet<string> GetDefaultVendorLocations()
                 return existingDirectories;
             }
 
-            throw new NotImplementedException();
+            // empty list for not implemented (e.g. Linux)
+            return new HashSet<string>();
```
benoithudson, line 766: I believe it's `/usr/autodesk` on linux.

We threw the exception on purpose to force us to figure it out when we ported to another platform.

vkovec: not sure how to check this works unless I have a linux machine I can install Maya onto

benoithudson: Let's add /usr/autodesk for linux, we can fix it later if needed.

Ideally the code would be:
* set `defaults` based on the platform
* do the loop filtering for the directory existing on all platforms the same way

com.unity.formats.fbx/Editor/ExportSettings.cs
```diff
@@ -1269,24 +1271,26 @@ private static string GetMayaExePathFromLocation(string location)
...
-                        return location + "/Contents/MacOS/Maya";
+                        return $"{location}/Contents/MacOS/Maya";
                     }
                     else
                     {
-                        return location + "/Maya.app/Contents/MacOS/Maya";
+                        return $"{location}/Maya.app/Contents/MacOS/Maya";
                     }
+                case RuntimePlatform.LinuxEditor:
+                    return $"{location}/Maya";
```
vkovec, line 1293: not sure about this

com.unity.formats.fbx/Editor/ExportSettings.cs
```diff
@@ -735,34 +735,34 @@ private static HashSet<string> GetCustomVendorLocations()
 
         private static HashSet<string> GetDefaultVendorLocations()
         {
-            if (Application.platform == RuntimePlatform.WindowsEditor)
+            HashSet<string> platformDefaults;
```
benoithudson, line 738: Looks like this could actually be just a string.

com.unity.formats.fbx/Editor/ExportSettings.cs
```diff
@@ -1183,17 +1183,19 @@ private static void FindDCCInstalls()
             var dccOptionNames = instance.dccOptionNames;
             var dccOptionPaths = instance.dccOptionPaths;
 
+            var vendorLocations = DCCVendorLocations;
+
             // find dcc installation from vendor locations
-            for (int i = 0; i < DCCVendorLocations.Count; i++)
+            for (int i = 0; i < vendorLocations.Count; i++)
```
benoithudson, line 1189: can this be `foreach (var vendorLocation in DCCVendorLocations)` ?

benoithudson (approved):
Some minor changes can be applied if you feel like it.

PR #674, opened 2022-10-18
Update master from release 5.0

vkovec:


benoithudson (approved):

vkovec:
it seems these are already in master after all

PR #676, opened 2022-10-21
add null check for sourceTransform

vkovec:


com.unity.formats.fbx/Editor/FbxExporter.cs
```diff
@@ -1735,6 +1735,11 @@ private List<ExpConstraintSource> GetConstraintSources(IConstraint unityConstrai
             unityConstraint.GetSources(sources);
             foreach (var source in sources)
             {
+                if (!source.sourceTransform)
```
benoithudson, line 1738: should we also check `!source` ?

benoithudson (approved):

PR #682, opened 2022-11-23
Remove trunk from other triggers for 4.2 branch

windxu88:
Remove trunk from other places for 4.2 branch


windxu88:
All the failures are because new `com.autodesk.fbx` package cannot be found which is expected at the moment.

vkovec (approved):

PR #686, opened 2022-12-15
FBX-382 expose more camera properties

vkovec:


mfe:
Tested on 2023.2.0a10: set a focus distance on a cam, exported as FBX, final FBX has the corresponding value when using 5.1.0-pre.1 and not in 5.0.0.
Also tested to import back the generated FBX, value is the right one in the inspector.

So LGTM

mfe (approved):

PR #693, opened 2023-02-17
FBX-457 replace LightType.Area with LightType.Rectangle

vkovec:


windxu88:
@vkovec My only concern is whether we need to add a guard such as `#if UNITY_2023_2_OR_NEWER` for the new API. Maybe no, because I checked the file history, `LightType.Area` and `LightType.Rectangle` coexist before the deprecate change. So `LightType.Rectangle` is still accessible for old Unity version such as 2023.1.

benoithudson (approved):

PR #700, opened 2023-05-02
Update doc for 5.1.0 (FBX-474)

mfe:
Just mentioning this new one.
 Should I mention somewhere we use depth of field property ?

### DIFF
--- a/com.unity.formats.fbx/Documentation~/exporting.md
+++ b/com.unity.formats.fbx/Documentation~/exporting.md
@@ -73,6 +73,7 @@ The FBX Exporter exports Physical Cameras, including these properties:
 
 - **Focal Length**
 - **Lens Shift**
+- **Focus Distance**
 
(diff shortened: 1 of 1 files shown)


vkovec (approved):
>Should I mention somewhere we use depth of field property ?

I think it's ok not to mention since it's not an option in Unity. IIRC I enabled it because it goes together with focus distance.

VertofChest:
Hey so quick Q
Does this work for assets packed into **.asset** files? I want to export them as .fbx if possible so I can tweak around with them.

Thanks!

mfe:
Hey @VertofChest, 

This is a code Pull Request so far from being the optimal way to ask question.
I suggest you ask on the Unity Forum the next time.

But yes, the FBX Exporter package can be used to export .asset files. 

I'm going to merge this PR now :)

PR #712, opened 2023-11-02
Add package field to .buginfo

alex-gubernsky:
## Purpose of this PR
As part of .buginfo adoption initiative and in accordance with the new rules for .buginfo files, this PR adds the `package` field into the `.buginfo` file. The field value corresponds to the Jira Package Insight object: `FBX Exporter`


mfe (approved):
LGTM

PR #713, opened 2024-01-11
CI: Update bokken images for release 4.1 branch

windxu88:
## Purpose of this PR:
Update deprecated bokken images for release 4.1 branch. 
Currently release/4.1 branch is still using old bokken images such as `upm-ci/win10:stable`. Because of that we can see following errors in CI:
```
Please switch to 'package-ci/win10:v4' or another supported Bokken image (note that package-ci/*:stable is deprecated and should no longer be used).
```
Besides, I also see CI failures caused by UPM-CI because of the following error:
```
Could not use "nc", falling back to slower node.js method for sync requests.
```

Updating deprecated bokken images seems to solve the issues above.

**JIRA ticket:**
[FBX-497](https://jira.unity3d.com/browse/FBX-497) CI: Update bokken images for release/4.1

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -99,9 +99,6 @@ nightly_test_trigger:
   name: Nightly tests Trigger
   triggers:
     recurring:
-      - branch: wind/update_bokken_image_release_4.1
-        frequency: daily
-        rerun: always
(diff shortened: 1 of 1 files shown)


mfe (approved):
LGTM

PR #723, opened 2024-08-13
Fixed exception when exporting with an LOD containing a null renderer (FBX-344)

mfe:
Jira issue: https://jira.unity3d.com/browse/FBX-344

It was raised by Viktoria a while ago. Not sure how she bumped into it but after a quick search, it is indeed possible to have an LOD with a null renderer. Cf issue for more details.

Note I also upgraded the version of the package because it seems it is the workflow APV 2.0 wants us to have (aka as soon as you add an Unreleased section, version should be updated).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changes in Fbx Exporter
 
+## [Unreleased]
+- Fixed an exception occurring during hierarchy export when an LOD contains a null renderer.
+
 ## [5.1.1] - 2024-03-18
 ### Changed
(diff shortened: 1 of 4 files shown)


lassond (approved):
Looks good. I wonder how the null renderer occurs in real-life scenarios. In any case, checking for null seems like the right thing to do here

mfe:
Yes I don't know what was the real use case indeed. But this can happen for example if the renderer is deleted.

thomas-tu (approved):

